PR #2, opened 2021-12-18
ephemeral_tabs.lua: Make ephemeral only when open via treeview and projectsearch

Jipok:


takase1121:
In general using `debug` isn't a very good practice.

Jipok:
> In general using debug isn't a very good practice.

It is even worse practice to follow the advice unconditionally without understanding it)
It would be nice if arguments were provided. Or an alternative solution without redesigning the core architecture with increasing code complexity.

Guldoman:
> It would be nice if arguments were provided.

Wouldn't this break if another plugin overrides `RootView:open_doc` before this one?

> Or an alternative solution without redesigning the core architecture with increasing code complexity.

https://github.com/lite-xl/lite-plugins/pull/106#issuecomment-991922609

Jipok:
> Wouldn't this break if another plugin overrides RootView:open_doc before this one?
> lite-xl/lite-plugins#106 (comment)

Well, your version of the code will most likely be. And I see no way to solve this.
By the way, I thought about this problem earlier. I propose to add to the core:
```lua
local function is_called_by(module)
  local i = 1
  while debug.getinfo(i) do
    if debug.getinfo(i).source:match(module .. "%.lua$") then
      return true
    end
    i = i + 1
  end
  return false
end
```

Jipok:
I noticed a problem with workspaces. After starting the editor, if you click in treeview on a file that is already open, it will become ephemeral. The problem arises because the session recovery mechanism does not call `RootView:open_doc`.
And in general, it does not allow you to save your information for plugins.

takase1121:
lol, your brain is looking for such a complex solution that you've missed the simplest one:

At https://github.com/lite-xl/lite-xl/blob/4be8a8b5821e8e34979d8580e71dbc08153986da/data/plugins/treeview.lua#L243:
```lua
core.root_view:open_doc(core.open_doc(doc_filename)).ephemeral
```

and yet you call me "worse practice to follow the advice unconditionally without understanding it"!

All is well until everyone decides to call `is_called_by` rapidly. Debug functions tend to be slow, so things slow down.
Oh? How about filename changes? Debug function **requires context**. You cannot expect to use debug functions and have it work all the time.

Jipok:
> lol, your brain is looking for such a complex solution that you've missed the simplest one
At https://github.com/lite-xl/lite-xl/blob/4be8a8b5821e8e34979d8580e71dbc08153986da/data/plugins/treeview.lua#L243:
`core.root_view:open_doc(core.open_doc(doc_filename)).ephemeral`

I don’t understand what you are suggesting. 

> Debug functions tend to be slow, so things slow down.

I don't see anything slow in [implementation](https://github.com/lua/lua/blob/eadd8c7178c79c814ecca9652973a9b9dd4cc71b/ldblib.c#L149). I did a check and called `is_called_by` 10000 times. Doing this was noticeable, but still faster than opening the project's changelog.md with syntax highlighting. 

> All is well until everyone decides to call is_called_by rapidly.

I would understand your concern if the call was in `draw` or something similar. But I only call this function twice in a rarely called single method. 

> How about filename changes?

Nice test that I haven't reviewed before. But everything works as I expected. Did you find any problem with this?

> Debug function requires context. You cannot expect to use debug functions and have it work all the time.

Well, I expect this to always work. The stack is not infinite, so the loop will end sooner or later, since getinfo is guaranteed to return nil. I don’t see any problems with this.

plugins/ephemeral_tabs.lua
```diff
@@ -6,17 +6,35 @@ local DocView = require "core.docview"
 local Doc = require "core.doc"
 local TreeView = require "plugins.treeview"
 
+local function is_called_by(module)
+  local i = 1
+  while debug.getinfo(i) do
+    if debug.getinfo(i).source:match(module .. "%.lua$") then
```
Guldoman, line 12: If you're dead set on using the debug function, at least avoid double calling and use the selector:
```lua
debug.getinfo(i, "S")
```
This will at least increase performance.

Guldoman: I still recommend to find another way, as using `debug.getinfo` is not a great idea.

Jipok: > If you're dead set on using the debug function

Suggest another solution.

Guldoman: I haven't tested it, but what was wrong with the solution I proposed before?

Jipok: Do you mean setting a global flag on a mouse event? In my opinion, this is a crutch.
We have a call stack. Consider that these are ready-made global flags for each function. What is the problem with using it?

Guldoman: Are you sure using a _local_ variable is worse that bothering the debug library?

Jipok: I'm not sure if "worse" is the right word here. Imagine if something like match.round was implemented in lua, and you offered me to use your own implementation for a small range of numbers. Well it looks like this to me.

Guldoman: It feels more like over-complicating a simple solution...

Jipok:
@Guldoman, any comments on closing?

Guldoman:
Look at the linked commit. It mentions this PR by chance, so github just decided to close this automatically.

TorchedSammy:
is there any chance for this to be completed/reworked cuz i'd personally like this

adamharrison:
Going through and merging things; I agree with 2021 @takase1121 about the debug thing.

Seems to me this could be pretty easily solved if we simply provide `TreeView.open`, and we simply override. Same with `projectsearch.ResultsView.open_selected_result`; would keep things more in line with the rest of the project, and is like a two line change. We should have it any way for hook reasons (and to reduce code duplication). I'll PR to lite.

adamharrison:
Closed in favour of #211 .

PR #4, opened 2021-12-20
colorpreview.lua: support alpha

Jipok:


jgmdev:
This works nicely but while typing I was able to crash it once and I don't remember what I wrote:

> Error: ~/.config/lite-xl/plugins/colorpreview.lua:25: attempt to perform arithmetic on a nil value (local 'a')

It seems something couldn't get converted to a number haven't, been able to reproduce it again, in any case if [tonumber()](https://www.lua.org/manual/5.4/manual.html#pdf-tonumber) fails to convert a value to a number it returns nil so maybe this:

```lua
if (a ~= nil) and (a ~= "") then
  a = tonumber(a, base)
  if base ~= 16 then
    a = a * 0xff
  end
else
  a = 0xff
end
```
should be changed to:
```lua
a = tonumber(a or "", base)
if a ~= nil then
  if base ~= 16 then
    a = a * 0xff
  end
else
  a = 0xff
end
```

PR #5, opened 2021-12-20
nonicons.lua: Use features from v2.1, icons in tabs

Jipok:
Uses [#104](https://github.com/lite-xl/lite-plugins/pull/104)

### DIFF
--- a/plugins/autoinsert.lua
+++ b/plugins/autoinsert.lua
@@ -1,4 +1,4 @@
--- mod-version:2 -- lite-xl 2.0
+-- mod-version:3 --lite-xl 2.1
 local core = require "core"
 local translate = require "core.doc.translate"
 local config = require "core.config"
--- a/plugins/autoinsert.lua
+++ b/plugins/autoinsert.lua
@@ -2,19 +2,20 @@
 local core = require "core"
 local translate = require "core.doc.translate"
 local config = require "core.config"
+local common = require "core.common"
 local DocView = require "core.docview"
 local command = require "core.command"
 local keymap = require "core.keymap"
 
 
-config.plugins.autoinsert = { map = {
+config.plugins.autoinsert = common.merge({ map = {
   ["["] = "]",
   ["{"] = "}",
   ["("] = ")",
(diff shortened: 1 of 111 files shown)


Jipok:
Example of icons in tabs:
image

The method with removing characters from the name is not very good. It won't break anything, but it still feels like a crutch.

Guldoman:
> The method with removing characters from the name is not very good. It won't break anything, but it still feels like a crutch.

We should probably split the `Node:draw_tab` function like we did for `TreeView:draw`.

Jipok:
> We should probably split the Node:draw_tab function like we did for TreeView:draw.

Doesn't make much sense. The problem is not the separation, but the arguments. Only the string that should be drawed is transmitted, which has already been modified by the plugins and the core. If `view` were passed, then we could try to get the true name of the file.
I could also iterate over all the `views` of the current `node`, and then compare the `get_name` to get the view of the tab we are drawing. But I don't like this solution even more.

Guldoman:
> The problem is not the separation, but the arguments.

Ah I see what you mean, I thought you were referring to the need of having to add spaces to move the text to the side.

>  If `view` were passed, then we could try to get the true name of the file.
> **UPD**: I will prepare the change so that the `view` is passed and not the title.

Yes, this should be good, and could allow plugins to do more interesting things.

I still think that splitting the tab drawing function would be useful.
For example the ephemeral tabs plugin could override the text drawing part to add the tildes, without having to change the name returned from the file.

Jipok:
The latest commit requires: https://github.com/lite-xl/lite-xl/pull/755
In general, everything seems to be ready. If you look closely at the pixel-by-pixel margins, it may seem to you that the icons are shifted to the left. But actually the problem with the close icon is:
image

jgmdev:
Thanks for your contributions, this plugin rocks! Merged into 2,1 branch with common.merge adaptation for config https://github.com/lite-xl/lite-xl-plugins/commit/31f0d678cf8972e816440b02e257fbb58a297626 so closing this PR

PR #8, opened 2021-12-22
add CI and script to reorganize README

takase1121:
This PR introduces a script to parse markdown to apply some transformations, namely:

- sort table entries in alphabetical order
- remove entries that are 404 (links that fail)

This script will absolutely destroy the table style, but it formats it properly afterwards.

Before this is merged, plugins that no longer exist on the repo should be removed first.

Here's a picture of a PR:
image

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,0 +1,32 @@
+name: Reorganize README
+
+on:
+  push:
+    branches: [master]
+
+jobs:
+  reorganize:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


takase1121:
The script could also be used to generate JSON files so that plugin managers like [slpm] and [lpm] can avoid parsing markdown themselves. This'll greatly simplify automation.

[slpm]: https://github.com/Nightwing13/slpm
[lpm]: https://github.com/debarchito/lpm

adamharrison:
Does this automated script create PRs when there are no changes to be made? I can only assume no (I'd think github wouldn't let you create the PR if there's no changes to commit); but just want to double-check.

takase1121:
No, I don't think so.

PR #10, opened 2021-12-24
fix: [sort] func passed to Doc:replace must have two return values

SpotlightKid:
Caused error `/usr/share/lite-xl/core/doc/init.lua:443: attempt to perform arithmetic on a nil value at /usr/share/lite-xl/core/init.lua:1114` when using `sort:sort`.


adamharrison:
Yup, looks good.

PR #11, opened 2021-12-24
feat: [sort] add alphabetic descending and natural sorting

SpotlightKid:
Natural sorting function implemented after http://notebook.kulchenko.com/algorithms/alphanumeric-natural-sorting-for-humans-in-lua (version four).


adamharrison:
I like it! The only thing I can think of is that it may make sense to symlink `sort:sort` to `sort:sort-alpha`; just so that it continues to work for people that've bound this to a key already; though this isn't critical (I can't imagine too many people have this on hotkey).

SpotlightKid:
Yes, I thought about that too. Would you suggest simply binding `function() do_sort(sort_alpha) end` to `[sort:sort]` as well as `[sort:sort-alpha]`?

adamharrison:
That makes sense to me.

SpotlightKid:
Closing this since I'm cleaning up my unmerged/unfinisged PRs and this is now conflicting and I don't use lite-xl anymore.

PR #16, opened 2022-01-01
RFC: export data

takase1121:
This PR aims to extract README.md content to make a JSON file that contains metadata for machine usage.

The format is as such:
```json
{
    "name": {
        "name": "name",
        "url": "url",
        "description": "description without markdown",
        "rawDescription": "description with markdown",
        "external": true,
    }
}
```

This is a RFC because there are a couple options how we can structure that JSON file (or not even JSON at all). A few options are:

1. CSV (easy to parse, but may require quoting)
2. TSV (easy to parse, less quoting since tabs are rare in README)
3. JSON (other format than as described above)
4. newline-delimited JSON (this would allow "streaming" parser by parsing each entry as it comes by)

**Note: I did mess up the branch significantly, I will mop up once I have free time.**


Jan200101:
Here are some extra attributes that I think are worth considering:
- `required_plugins` a list of plugins required to function
- `hidden` tells a plugin managers to not show this plugin, useful for plugins that intend to be a base for something
- `{max,min}_version` To specify on what version of Lite-XL the plugin can run
- `license`
- `checksum` useful to ensure the plugin is the one expected (though this can perhaps be offloaded into something else)

I don't think lite-xl-plugins should act as a centralised plugin repository, if we choose to support plugin managers, and instead move plugins into dedicated folders with a file to specify this information (`info.json`).
That way things like updating plugins can be done using version control like git as well as allow independent developers to create plugin repositories.

Is there a need for a rich description field? I think it would be better to simply keep the rawDescriptioon field and rename it to description to keep the dependencies on what is needed to display information down

takase1121:
- `required_plugins` this alone would be impossible to implement "correctly". Perhaps one could trace `require`, but that's probably the most you can do.
- `hidden` there's no indication of this on README, so I don't plan to add.
- `{max,min}-version` this can be considered but plugins usually only is compatible to 1 version because just how much API change there is
- `license` this one is icky. We'll need to determine if all the plugins here uses some sort of license.
- `checksum` GitHub API has [SHA](https://stackoverflow.com/a/20208173)

I understand that there's a lot to be desired - versioning, more metadata and something, but I think that a lot of work is redundant. I'd much rather work on [LuaRocks integration](https://github.com/lite-xl/lite-xl/discussions/751). Everyone here is eager to re-invent the wheel with this repo, but I think its honestly not worth it.

PR #18, opened 2022-01-02
Convert Plugins and Colours into JSON File

Nightwing13:
# About
This python script is supposed to be an improvement over @takase1121 's script. It hashes plugins websites so that features like updates can be implemented within plugin manager

# Purpose
Anyone can reimplement it in their favourite language. (Since I know, even though my script works, by any means, it is not a good implementation(Since I am not versed in python))
It can be used by package managers and the best part is it tracks changes. (By hashing plugins(websites or files) when a PR is merged)

# Limitations
I don't know how to make a CL script(Github Actions) for it(Since I cannot modify existing files (I like to see something like Takase did for Readme.md)).

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -30,6 +30,7 @@ jobs:
           cd ..
           git clone --recursive ${{ secrets.ssh }}
           mv scripts/extensions_tmp.json lite-xl-plugins/scripts/extensions.json
+          cd lite-xl-plugins/
           git add .
           git commit -m "update extensions"
           git push -u origin dev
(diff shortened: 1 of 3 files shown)


takase1121:
The commit history can be accessed with GitHub API. This is redundant.

Nightwing13:
> The commit history can be accessed with GitHub API. This is redundant.
Can it be used for accessing the commit history of individual files? (For example, If I want to access something like autowrap.lua)

takase1121:
> > The commit history can be accessed with GitHub API. This is redundant.
> 
> Can it be used for accessing the commit history of individual files? (For example, If I want to access something like autowrap.lua and also, I won't be cloning the repo)



> > The commit history can be accessed with GitHub API. This is redundant.
> 
> Can it be used for accessing the commit history of individual files? (For example, If I want to access something like autowrap.lua and also, I won't be cloning the repo)

https://stackoverflow.com/a/15838387

PR #20, opened 2022-01-16
adding gemtext highlighting plugin

Not-a-web-Developer:
I heavily modified the markdown plugin to work with gemtext, a markup language for the [Gemini Protocol][1], which is very similar to markdown.

this is what the highlighting looks like in default theme:
image

[1]: https://gemini.circumlunar.space


plugins/language_gmi.lua
```diff
@@ -0,0 +1,22 @@
...
+local syntax = require "core.syntax"
+
+
+
+syntax.add {
+  name = "Gemtext",
+  files = { "%.gmi$" },
+  patterns = {
+    { pattern = { "```", "```" },           type = "string"   },
+    { pattern = "#.-",                      type = "keyword"  },
```
Guldoman, line 11: If you want to highlight the whole line, you should use `#.*`.
Otherwise, if you only want to highlight the `#`, you should use `#+`

plugins/language_gmi.lua
```diff
@@ -0,0 +1,22 @@
...
+    { pattern = "%*.*",                     type = "keyword2" },
```
Guldoman, line 12: This is just my preference, but I think that we should only highlight the `*`, not the whole line.

plugins/language_gmi.lua
```diff
@@ -0,0 +1,22 @@
...
+    { pattern = "=>",                       type = "function" },    
+    { pattern = "!?%[.-%]%(.-%)",           type = "function" },
```
Guldoman, line 14: Does GemText support this style of links?

Guldoman (commented):
Does Gemtext allow spaces before `#`, `*`, `>` and `=>`?
If it does we should add `%s*` in front of those patterns.

plugins/language_gmi.lua
Not-a-web-Developer: Alright, I had just modified what was in the markdown plugin; I'll fix it here and fix it in the markdown plugin too.

plugins/language_gmi.lua
Not-a-web-Developer: my thinking behind highlighting the entire line for bullet lists was for it to match how headings and quotes are handled (the entire line is highlighted in those cases), but highlighting just the `*` does improve readability, so i'll change that accordingly.

plugins/language_gmi.lua
Not-a-web-Developer: i forgot to remove line 14 (gemtext doesn't support that style of links), i'll remove that asap.

Not-a-web-Developer:
> Does Gemtext allow spaces before `#`, `*`, `>` and `=>`? If it does we should add `%s*` in front of those patterns.

gemtext doesn't allow spaces before the symbols, so adding that wouldn't be necessary.

Not-a-web-Developer:
I've done all the changes, please tell me if more are needed.

PR #27, opened 2022-01-25
Add debugger

Jan200101:
Upstream https://github.com/slembcke/debugger.lua
   with changes to allow for breakpoints for better debugging


Jan200101:
Took me a bit to get this over the finish line

This is a debugger based on [debugger.lua](https://github.com/slembcke/debugger.lua) with capabilities added simplistic breakpoints and optional logic that depends on external luarocks removed.

To invoke it you have to start Lite-XL in a console and run the `debugger:break` command, which will trigger a breakpoint.

From there on you have an interactive REPL in the console with a heap of functionalities.

You can also set breakpoints on C functions, but you cannot step inside them.

No more print debugging!

PR #28, opened 2022-01-26
add plugin select_colorscheme

AlexSol:


franko (approved):
Fine for me, thank you

plugins/select_colorscheme.lua
```diff
@@ -0,0 +1,135 @@
...
+    if v == value then
+      remove_idx = i
+      break
+    end
+  end
+  if remove_idx then
+    table.remove(list, remove_idx)
+  end
+  return list
+end
```
franko, line 105: That's fine but I proposed a shorter, simplified version. There is a reason to keep it that way?

franko (approved):
Just a question about one function that can be simplified but that's fine for me.

plugins/select_colorscheme.lua
AlexSol: > That's fine but I proposed a shorter, simplified version. There is a reason to keep it that way?

I agree with you. my mistake

franko:
Great plugin, thank you for the contribution and for taking into account my suggestions.

PR #29, opened 2022-01-27
Plugin to improve core:find-file

jgmdev:
This plugin overrides the `core:find-file` command in order to better support searching of files on large project trees. It generates a full project files index by using the `find` command on unix and the `dir` command on windows. It calls this commands using the `process` api to eliminate or prevent gui stalls. The project files index is cached and only updated when a `directory on change event` is received or the user changes the project directory. Haven't tested on Windows but here is a video testing it on the linux kernel source tree:

https://user-images.githubusercontent.com/1702572/151286060-06ceb20a-d71f-4981-b8c9-0c87cd056cbd.mp4


jgmdev:
Some changes:
* Renamed plugin to findfileimproved.lua. 
* Instead of relying on system command utilities (find, dir) to recurse project files made use of `system.list_dir` which has similar performance and should work on any OS.
* Using system.list resulted in simplification of the plugin code.

jgmdev:
Closing this look at https://github.com/jgmdev/lite-xl-threads instead

PR #30, opened 2022-01-29
Remove `themeselect` in favor of `select_colorscheme`

BenStigsen:
I'm removing my `themeselect` in favor of `select_colorscheme`


Guldoman:
Don't they serve different purposes?
If I understand correctly `themeselect` changes the theme based on the current file, while `select_colorscheme` is used as an easy way to select the theme the editor uses.

BenStigsen:
I'll be completely honest, I actually think that I'm brainfarting. `themeselect` does not look like my code or plugin. But I did create a plugin by that name, maybe I never uploaded it... I'm so lost now. But if they serve different purposes which it seems they do, then let's keep it.

PR #31, opened 2022-02-08
indentguide: use indent size from doc instead of config

TorchedSammy:
personally i prefer using size 4 tabs for indentation, but the lite xl codebase and plugins use 2 spaces which can cause it to not be aligned properly


Guldoman:
This won't work in the current release as `doc:get_indent_info` is only present in `master`.
When 2.1 will be released this could be merged in.

TorchedSammy:
gonna close and make a new pr from a dedicated branch

PR #32, opened 2022-02-09
Improvements on syntax of sh(using bash keywords)

dheisom:
New keywords on shell script syntax(from `help` builtin command of bash)

Old: 
old

New:
new


jgmdev:
Looks good, thanks.

PR #35, opened 2022-02-16
Added language_pkgbuild

jgmdev:
Adds support for [PKGBUILD](https://wiki.archlinux.org/title/PKGBUILD) files used for writing packaging build instructions.


Guldoman:
I wonder if this could have been done by "importing" `language_sh` and building on it. This would have meant having it as a dependency which could be undesirable.

Is there anything that could be backported to `language_sh`?

jgmdev:
> Is there anything that could be backported to `language_sh`?

I will later work on language_sh,  stuff which comes to mind to backport are the improved patterns for numbers, command arguments, support for variable declarations and additional keywords for standard tools like mkdir, mv, cp, etc...

PR #36, opened 2022-02-19
add treeview-menu-extender plugin

juliardi:


jgmdev:
Duplicate, I was needing that the other day and did a Save As, also as far as I recall with `Rename` you can do the same as `Move to` by specifying a different path when doing the rename.

juliardi:
1. Of course you can duplicate a file by using Save As. But, I think it would be easier if there is a menu we can click in the treeview.
2. Yes, `Move To` uses `Rename` function under the hood, because it achieves the same thing. However, `Rename` only displays filename and not everyone would think that they can move a file using `Rename` by specifying a different path. 

This plugin does not add a significant functional to Lite-XL, but IMO it gives better UX in terms of file operation.

jgmdev:
Was testing before a merge duplicate worked perfectly but got this error when trying a "Move To"

```
Fri Mar  4 15:10:31 2022 [ERROR] ~/.config/lite-xl/plugins/treeview-extender.lua:111: attempt to index a nil value (field 'hovered_item') at /usr/share/lite-xl/core/init.lua:1366
stack traceback:
~/.config/lite-xl/plugins/treeview-extender.lua:111: in field 'predicate'
/usr/share/lite-xl/core/contextmenu.lua:67: in function 'core.contextmenu.show'
/usr/share/lite-xl/core/contextmenu.lua:185: in function 'core.contextmenu.on_mouse_pressed'
/usr/share/lite-xl/plugins/treeview.lua:444: in function 'core.rootview.on_view_mouse_pressed'
/usr/share/lite-xl/core/rootview.lua:137: in function 'core.rootview.on_mouse_pressed'
/usr/share/lite-xl/core/init.lua:1467: in upvalue 'on_event'
/usr/share/lite-xl/plugins/macro.lua:19: in function 'core.on_event'
[C]: in function 'xpcall'
/usr/share/lite-xl/core/init.lua:1365: in function 'core.try'
/usr/share/lite-xl/core/init.lua:1525: in function 'core.step'
/usr/share/lite-xl/core/init.lua:1602: in function 'core.run'
(...tail calls...)
[string "local core..."]:9: in function <[string "local core..."]:2>
[C]: in function 'xpcall'
[string "local core..."]:2: in main chunk
```

juliardi:
The error seems to happen when the treeview lost focus, thus the
`hovered_item` becomes null. I'll see what I can do to fix it.

juliardi:
@jgmdev I have updated the plugin, can you please check it again?

jgmdev:
Sorry took me so long to try latest changes, seems to be working nicely even on master branch of lite-xl, merging.

PR #42, opened 2022-03-08
Made a plugin

ThaCuber:
Just wanted to add it to the repo. It's an equation grapher, it graphs y=x type equations.


jgmdev:
Interesting stuff, thanks!

PR #43, opened 2022-03-16
add `language_glaz.lua`

StunxFS:


plugins/language_glaz.lua
```diff
@@ -0,0 +1,100 @@
+-- mod-version:2 -- lite-xl 2.0
+
+local syntax = require "core.syntax"
+
+syntax.add {
+    name = "Glaz",
+    files = {"%.glaz$", "%glaz.proj$"},
```
Guldoman, line 7: In the second pattern, the `%` is needed before `.`, and not at the beginning.

Also, this will match files like `blablaglaz.proj`, is this intended? If you want to match only files named exactly `glax.proj` you should use `^glaz%.proj$`.

Guldoman (changes requested):
After fixing the second file pattern I think that we can merge this.

Thank you for your contribution!

PR #47, opened 2022-03-22
Added typingspeed plugin

B14CK313:
This plugin displays your current typing speed in characters per minute
and words per minute in the status bar.
I tried to keep the logic very simple, so instead of using real accurate
values, it counts characters/words in the current minute and adds
the value of the previous minute multiplied by the percentage needed to
'complete' the current minute (so it basically always displays the value
for one minute, but the part of it that falls into the last minute uses
the average):
```
words_last * (1 - (t.sec) / 60) + words
```
This should give fairly good results without the need for complicated code
or falling back to 0 after every minute.

It should be compatible with both the 2.0.5 release and the new config and status_view APIs.


jgmdev:
Tested and seems to work nicely, ty!

PR #48, opened 2022-03-23
update `language_glaz`

StunxFS:


StunxFS:
The `Self` keyword has the same behavior as Rust's `Self` in my language :).﻿

PR #49, opened 2022-03-25
Added Native Plugin Auto Compilation

adamharrison:
Allows auto-compilation of native plugins for lite, if your system has a C compiler installed. Currently, only targets GCC, would be easy to add LLVM.


plugins/support_cc.lua
```diff
@@ -0,0 +1,54 @@
...
+    if stderr_chunk ~= nil then stderr = stderr .. stderr_chunk end
+  end
+  local stdout_chunk, stderr_chunk = process:read_stdout(), process:read_stderr()
+  if stdout_chunk ~= nil then stdout = stdout .. stdout_chunk end
+  if stderr_chunk ~= nil then stderr = stderr .. stderr_chunk end
+  return process:returncode(), stdout, stderr
+end
+
+function cc.compile_plugin(target, options)
+  local dynamic_suffix = PLATFORM == "Mac OS X" and 'lib' or (PLATFORM == "Windows" and 'dll' or 'so')
```
Jan200101, line 26: isn't it dylib on Mac OS?

adamharrison: You're right, it is. Weird; I just copied this from start.lua; we'll want to change that there too.

plugins/support_cc.lua
```diff
@@ -0,0 +1,54 @@
+-- mod-version:3 --lite-xl 2.1
+
+local core = require "core"
+local config = require "core.config"
+local common = require "core.common"
+
+local cc = common.merge({
+  backend = "gcc"
```
Jan200101, line 8: Should we really differenciate between gcc and clang? clang is a drop in solution in most cases.

plugins/support_cc.lua
```diff
@@ -0,0 +1,54 @@
...
+      local info = system.get_file_info(directory .. PATHSEP .. v)
+      if info and info.modified > lib_info.modified then
+        requires_compile = true
+        break
+      end
+    end
+  end
+  if requires_compile then
+    if cc.backend == "gcc" then
+      local cmd = { "gcc", "-fPIC", "-shared", "-g", "-o", path, "-I" .. EXEDIR .. PATHSEP .. "resources" }
```
Jan200101, line 43: If we drop compiler specific backends we should simply invoke `/usr/bin/cc` since that is symlinked to the system default C compiler.

Reading the `CC` env var is also common practice, though perhaps that is better as a config value

adamharrison: I feel like you may want to sub a specific compiler; and the options are not generic to all compilers. Definitely we should look at CC env variable, though.

PR #52, opened 2022-03-30
Plugin Manager

adamharrison:
Threw in a plugin manager. Designed to be a base API, that someone can throw a View in front of, if they want, but by default has an extremely basic interface for install/unsinstall based on `CommandView`.

* Does dependency resolution by parsing `package_name.lua` or `package_name/init.lua`, and looking at package mentions. Once we have a method of versioning these (I'm thinking we just add it to the comment line), we should probably also have a `require_plugin` function which allows for a version specification, that this will then parse.
* Will automatically check against the local system and user plugin repositories for plugins before trying to download them. 

Will scan the README soon, and add in all the remote repositories and also scan them.

Currently depends on `support_curl`, which is a native package.


adamharrison:
Also, just a heads up; `support_cc`, `support_curl` are separate PRs who's commits I've just merged in here, because they're needed; but those should be merged before this.

dheisom:
Hello I am developing a plugin manager for lite-xl too, the repository is github.com/dheisom/lxpm

adamharrison:
Yeah; that's totally fine. This isn't part of the core; so it's not an "official" package manager in that sense; it's just a plugin, so totally cool with including both into the plugins repo, and people can use whichever one they like.

TorchedSammy:
> Does dependency resolution by parsing package_name.lua or package_name/init.lua, and looking at package mentions

a plugin manifest in my opinion is a way better solution with less headache. does it know the difference between a lite xl plugin and a generic lua module i have installed a different way?

adamharrison:
Closing, as this is now obsolete; will add in a link to the new plugin manager repo.

PR #56, opened 2022-04-02
Add `language_autohotkey.lua`

devPnal:
Add AutoHotkey plugin (language-autohotkey.lua)
Description: https://github.com/devPnal/lite-plugin-autohotkey

I had mistakes in previous PR, so I will close it and PR it again.


adamharrison:
I'm totally OK with this being a remote plugin if you want; but the language plugin is small enough we're happy to just include it directly in the repository.

Up to you; just confirm for me you do want it as a remote link, and I'll merge.

devPnal:
Of course, the language plugin is small enough to be included directly in the repository,
but AutoHotkey is still advancing, it needs to be modified immediately, so I want it as a remote link. (=Because I can fix it immediately)

Thank you for your kindness.

adamharrison:
Sure; sounds good. Merged.

PR #59, opened 2022-04-13
added mdpreview, a markdown previewer

Not-a-web-Developer:
since the `ghmarkdown` plugin was using the GitHub API for converting markdown to html, there were privacy concerns (let alone the fact that it wouldn't work offline); so I wrote a markdown previewer that has the same user experience as `ghmarkdown`, but uses either luamd (a markdown rendering library in lua) or pandoc to render markdown, depending on which branch is used.

was requested by @jgmdev and myself, and it's finally complete enough for me to publish it.


jgmdev:
Another idea: instead of having two branches why not try to use pandoc by default and if not found on PATH fallback to luamd, also you could add a config option called `renderer` (eg: config.plugins.mdpreview.renderer) or something similar which could have one of three string values:

* "auto" - check if pandoc is available and use it otherwise fallback to luamd
* "pandoc" - use pandoc by default and throw error message if not in PATH
* "luamd" - use luamd by default

Not-a-web-Developer:
> Another idea: instead of having two branches why not try to use pandoc by default and if not found on PATH fallback to luamd, also you could add a config option called `renderer` (eg: config.plugins.mdpreview.renderer) or something similar which could have one of three string values:
> 
> * "auto" - check if pandoc is available and use it otherwise fallback to luamd
> * "pandoc" - use pandoc by default and throw error message if not in PATH
> * "luamd" - use luamd by default

that's a good idea, i should probably change this to a draft PR to work on it.

SwissalpS:
I like the idea of an offline markdown preview.
Are you still working on this?

Issue #74, opened 2022-04-25
bracketmatch doesn't ignore comments


ShadiestGoat:
![image](https://user-images.githubusercontent.com/48590492/165003892-4f86de3e-f356-4dda-9e92-616f9e667e31.png)

PR #81, opened 2022-04-30
Add more `bracketmatch` styles

### LINKED ISSUE #73: bracketmatch highlight both pairs
There should be options for how highlighting is done. Should give options about highlighting rather than underlining, and whether or not to highlight the original bracket.

Guldoman:
Now `bracketmatch` ignores brackets found in comments and has a few different styles.

```lua
-- Colors can be configured as follows:
--   underline color  = `style.bracketmatch_color`
--   bracket color    = `style.bracketmatch_char_color`
--   background color = `style.bracketmatch_block_color`
--   frame color      = `style.bracketmatch_frame_color`

config.plugins.bracketmatch = {
  highligh_both = true, -- highlight the current bracket too
  style = "underline",  -- can be "underline", "block", "frame", "none"
  color_char = false,   -- color the bracket
  line_size = math.ceil(1 * SCALE), -- the size of the lines used in "underline" and "frame"
}
```
Underline
Screenshot from 2022-04-30 04-35-01
Underline + color_char
Screenshot from 2022-04-30 04-35-08
Block
Screenshot from 2022-04-30 04-35-23
Block + color_char
Screenshot from 2022-04-30 04-35-30
Frame
Screenshot from 2022-04-30 04-35-40
Frame + color_char
Screenshot from 2022-04-30 04-35-48
None + color_char
Screenshot from 2022-04-30 04-35-55

Closes #73 and #74.

### DIFF
--- a/plugins/bracketmatch.lua
+++ b/plugins/bracketmatch.lua
@@ -7,27 +7,34 @@ local DocView = require "core.docview"
 
 local bracket_maps = {
   -- [     ]    (     )    {      }
-  { [91] = 93, [40] = 41, [123] = 125, step =  1 },
+  { [91] = 93, [40] = 41, [123] = 125, direction =  1 },
   -- ]     [    )     (    }      {
-  { [93] = 91, [41] = 40, [125] = 123, step = -1 },
+  { [93] = 91, [41] = 40, [125] = 123, direction = -1 },
 }
 
+local function get_token_at(doc, line, col)
+  local column = 0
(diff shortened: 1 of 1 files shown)


jgmdev:
Since we are kind of getting ready for 2.1, can this get rebased for 2.1 instead? To be honest I feel kind of lazy rebasing it my self for 2.1 (since I would need to first understand the changes). Also the new config options seem like a good candidate for the settings gui plugin :)

jgmdev:
Merging since I already ported it to 2.1

PR #85, opened 2022-05-03
Added in clock plugin for statusview.

adamharrison:
Noticed this was missing (there's BigClock; but I want something smaller). Need this for write-xl.


adamharrison:
Whoops; smallclock.lua exists; will use that.

PR #97, opened 2022-06-09
`minimap`: Performance improvements

Guldoman:
With this PR:
* Rects position and color is cached.
* Touching rects with the same color are merged to avoid doing useless `draw_rect` calls.
* The `spaces_to_split` option is used to specify the minimum amount of spaces needed to split a token.
  This is 2 by default; use 1 to go back to previous behavior.
* Toggling the minimap works even if it's globally disabled.

The main disadvantage of the caching is that invalidation doesn't happen in some cases. For example after changing the color scheme.

Doing some tests I get an improvement of ~30% on the CPU usage while scrolling and a reduction of ~1000 `draw_rect` calls.

### DIFF
--- a/plugins/minimap.lua
+++ b/plugins/minimap.lua
@@ -5,7 +5,22 @@ local common = require "core.common"
 local config = require "core.config"
 local style = require "core.style"
 local DocView = require "core.docview"
+local Highlighter = require "core.doc.highlighter"
 local Object = require "core.object"
+
+-- cache for the location of the rects for each Doc
+local highlighter_cache
+local function reset_cache()
+  highlighter_cache = setmetatable({}, { __mode = "k" })
+end
+reset_cache()
+
+-- minimap status per DocView
+local per_docview
+local function reset_per_docview()
+  per_docview = setmetatable({}, { __mode = "k" })
+end
+reset_per_docview()
 
 -- Sample configurations:
 -- full width:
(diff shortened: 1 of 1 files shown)


jgmdev:
> The main disadvantage of the caching is that invalidation doesn't happen in some cases. For example after changing the color scheme.

on initial caching you can store a reference to any of the color tables in style.syntax like style.syntax["normal"] and if changed then invalidate the cache.

jgmdev:
> Yeah, but I'd have to store all the different colors that are used and compare them all at each update, which feels like a bit of an hack.

That assumption is wrong. Every time a new color scheme is loaded the table representing each of the colors is replaced, and when I say table I mean each of the color table ones like `style.syntax["normal"]` which holds rgba, so you only need to check for one of the colors in `style.syntax` to see if the table is different, for example:

```lua
local prev_color = style.syntax["normal"] -- holds a reference to the color table
--- ... do something
if prev_color ~= style.syntax["normal"] then
   invalidate_cache()
   -- update our reference
   prev_color = style.syntax["normal"] 
end
```

Remember that each color scheme doesn't overrides `style.syntax`, but the sub fields like:
```lua
style.syntax["normal"] = { common.color "#efdab9" }
```

Guldoman:
Yeah, but that works only assuming that every color scheme overrides `"normal"` (which seems to be the case at the moment, so it would be fine).

There is also the problem of a user changing the *inside* of a color which would not result in a table change.

In the end this is a reasonable check, so I'll implement this.

jgmdev (approved):

PR #98, opened 2022-06-11
Add support for SQL strings to language_php

KatrinaKitten:
Naively detects strings starting with `CREATE`, `SELECT`, `INSERT`, `UPDATE`, `REPLACE`, `DELETE`, `DROP`, or `ALTER` (case insensitive), and marks them as a subsyntax for `.sql` files.

Note that this doesn't create a hard dependency on `language_psql` or any other specific SQL highlighting plugin; if no such plugin is present, SQL strings will simply appear un-highlighted instead of as strings.

### DIFF
--- a/plugins/language_php.lua
+++ b/plugins/language_php.lua
@@ -10,6 +10,15 @@ local syntax = require "core.syntax"
 require "plugins.language_css"
 require "plugins.language_js"
 
+-- generate SQL string marker regex
+local sql_markers = { 'create', 'select', 'insert', 'update', 'replace', 'delete', 'drop', 'alter' }
+local sql_regex   = {}
+for _,marker in ipairs(sql_markers) do
+    table.insert(sql_regex, marker)
+    table.insert(sql_regex, string.upper(marker))
+end
+sql_regex = table.concat(sql_regex, '|')
+
 -- define the core php syntax coloring
 syntax.add {
   name = "PHP Source",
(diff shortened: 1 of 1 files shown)


Guldoman:
What do you think about using an approach like this:
```lua
local sql_words = {
  "create",
  "select",
  "insert",
  ...
}

local sql_combined = "("
for k,v in pairs(sql_words) do
  local multi_case = "(?:"
  for i=1,#v do
    local char = string.sub(v, i, i)
    multi_case = multi_case.."["..string.lower(char)..string.upper(char).."]"
  end
  sql_words[k] = multi_case..")"
end
sql_words = "(?:"..table.concat(sql_words, "|")..")"

...
    {
        regex = { '"(?='..sql_words..')', '"', '\\' },
        syntax  = '.sql',
        type    = "keyword"
    },
    {
        regex = { "'(?="..sql_words..")", "'", "\\" },
        syntax  = '.sql',
        type    = "keyword"
    },
...
```

This way you automatically generate all the "multi-cased" keywords and only need two regexes.

KatrinaKitten:
To be honest I'd forgotten since reading the docs that `regex` was an option or I absolutely would have. I did try a generative approach, but Lua patterns don't support alternation, and I wasn't having much luck with `table.unpack` to pregenerate all the entries. I'll make a few tweaks to implement that.

Guldoman:
Just to confirm, you're dropping the mixed case support, right?

KatrinaKitten:
Technically yes, it was initially a quirk of how pattern syntax required it to be written. I can add it back if that's preferred though.

Guldoman:
Is the first letter often capitalized? If not we can leave it as it is. If someone really wants to support mixed case, they can just open a PR.

I'd say just do what you think is best.

KatrinaKitten:
Official style guide is all caps. I see a lot of all lowercase, but rarely if ever mixed.

PR #101, opened 2022-06-12
language_v: add syntax support for raw strings

benwalksaway:
This PR will add raw strings support for the V programming language. (see [V strings](https://github.com/vlang/v/blob/master/doc/docs.md#strings))
Escape characters in a raw string are treated as escaped escape characters. (`\` == `\\` in a raw string)

**before**:
v_raw_string_no_support

**after**:
v_raw_string_support

### DIFF
--- a/plugins/language_v.lua
+++ b/plugins/language_v.lua
@@ -12,8 +12,8 @@ syntax.add {
     { pattern = { '"', '"', '\\' },         type = "string"   },
     { pattern = { "'", "'", '\\' },         type = "string"   },
     { pattern = { "`", "`", '\\' },         type = "string"   },
-    { pattern = "r()'.*'",                  type = {"normal", "string"} }, -- raw string single quotes
-    { pattern = "r()\".*\"",                type = {"normal", "string"} }, -- raw string double quotes
+    { pattern = "r()'.*'",                  type = { "normal", "string" } }, -- raw string single quotes
+    { pattern = "r()\".*\"",                type = { "normal", "string" } }, -- raw string double quotes
     { pattern = "0x[%da-fA-F_]+",           type = "number"   },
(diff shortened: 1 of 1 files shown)


plugins/language_v.lua
```diff
@@ -12,6 +12,8 @@ syntax.add {
     { pattern = { '"', '"', '\\' },         type = "string"   },
     { pattern = { "'", "'", '\\' },         type = "string"   },
     { pattern = { "`", "`", '\\' },         type = "string"   },
+    { pattern = "r()'.*'",                  type = { "normal", "string" } }, -- raw string single quotes
+    { pattern = "r()\".*\"",                type = { "normal", "string" } }, -- raw string double quotes
```
Guldoman, line 16: Do raw strings support multiline?
If so I'd use something like
```suggestion
    { pattern = { "r'", "'" },              type = "string" }, -- raw string single quotes
    { pattern = { 'r"', '"' },              type = "string" }, -- raw string double quotes
```

benwalksaway: Yes, V raw strings support multiline. But is there a way to exclude the `r` from the string highlighting? If not, I will commit your suggestion :slightly_smiling_face:

Guldoman: You could do something ~~ugly~~ like this:
```lua
    { pattern = { 'r%f["]', '%f["]' }, -- raw string double quotes
      syntax = {
        patterns = {
          { pattern = ".*", type = "string" },
        },
        symbols = {},
      },
      type = "normal"
    },
```
but other language plugins like python include the "letter" in the token https://github.com/lite-xl/lite-xl/blob/99d328cfd793420bd1890555330b71384616b290/data/plugins/language_python.lua#L11-L14

In the end just commit the version you prefer, either way is fine for me.

PR #102, opened 2022-06-13
Add language_htaccess

KatrinaKitten:
Adds the `language_htaccess` plugin , which supports syntax highlighting for `.htaccess` files.


plugins/language_htaccess.lua
```diff
@@ -0,0 +1,76 @@
...
+  symbols = {}
+}
+local xml_syntax = {
+  patterns = {{ pattern = { '"', '"', '\\' }, type = "string" }},
+  symbols = {}
+}
+
+syntax.add {
+  name = ".htaccess File",
+  files = { "%.htaccess$" },
```
Guldoman, line 40: Should this match `.htaccess` only, or can those files begin with something else (`example.htaccess`)?

If the former, I'd use:
```suggestion
  files = { "^%.htaccess$" },
```
If the latter, then it's correct as it is now.

KatrinaKitten: Fixed as of latest commit.

plugins/language_htaccess.lua
```diff
@@ -0,0 +1,212 @@
...
+    -- URLs
+    { pattern = { "%w-://", "%f[%s]" },         type = "string", syntax = url_syntax },
+    { pattern = { "%f[%S]/", "%f[%s]" },        type = "string", syntax = url_syntax },
+    -- Mime types
+    { pattern = "%f[%w]application/[%w%._+-]+", type = "keyword2" },
+    { pattern = "%f[%w]font/[%w%._+-]+",        type = "keyword2" },
+    { pattern = "%f[%w]image/[%w%._+-]+",       type = "keyword2" },
+    { pattern = "%f[%w]text/[%w%._+-]+",        type = "keyword2" },
+    { pattern = "%f[%w]audio/[%w%._+-]+",       type = "keyword2" },
+    { pattern = "%f[%w]video/[%w%._+-]+",       type = "keyword2" },
```
Guldoman, line 186: What's the reason for `%f[%w]`?

Guldoman (approved):
~~That's *a lot of* keywords wow~~

Other than a little clarification, this looks good to go. Thank you!

PR #103, opened 2022-06-14
[nonicons] Fixed wrong icon on opened directory

dheisom:
The icon of a opened directory was wrong, the unicode character that was in use don't exists anymore on the new version of nonicons font

### DIFF
--- a/plugins/nonicons.lua
+++ b/plugins/nonicons.lua
@@ -121,7 +121,7 @@ function TreeView:get_item_icon(item, active, hovered)
     font = icon_font
     color = style.text
     if item.type == "dir" then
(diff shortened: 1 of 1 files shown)


Guldoman:
Yeah, it looks like nonicons was updated.
The VSCode extension uses the same codepoints you're proposing
https://github.com/yamatsum/vscode-nonicons/blob/fbc889160bf39c5471516d7151633923cbfc9b14/icon-theme.json#L29-L36

We should probably link to the font download locked on a commit, like https://github.com/yamatsum/nonicons/raw/6a2faf4fbdfbe353c5ae6a496740ac4bfb6d0e74/dist/nonicons.ttf

PR #107, opened 2022-06-17
Syntax highlighting for intel x86 assembly

DMClVG:
Highlights intel syntax x86 instructions and NASM directives :)


Guldoman (commented):
That's *a lot* of symbols.
This looks good.
Would you mind adding support for hex numbers in the `%x+[hH]` format?

DMClVG:
Here you go. I've made some other small changes as well

DMClVG:
GTG

plugins/language_assembly_x86.lua
```diff
@@ -0,0 +1,2211 @@
+-- mod-version:2 -- lite-xl 2.0
+-- Support for the intel syntax x86-64 assembly
+-- Simply add it to lite-xl's plugins folder (located somewhere at home/.config/lite-xl/plugins)
+-- https://github.com/DMClVG
+
+local syntax = require "core.syntax"
+
+syntax.add {
+  files = { "%.asm$", "%.S$", "%.s$" },
```
Guldoman, line 9:
```suggestion
  files = { "%.asm$", "%.[sS]$" },
```

DMClVG:
> for this to work you would need to add into the `patterns` table a rule like:
> 
> ```lua
> { pattern = "%.%.%a+", type = "normal" }
> ```

Works but, how is normal different from symbol? They both seem to work here

Guldoman:
> Works but, how is normal different from symbol? They both seem to work here

The idea is that each token you match is *always* looked up in the `symbols` table.
So to make things work you need to provide a pattern that can match the keys in the table.
The token `type` of the pattern will be used if no match is found in the `symbols` table.

PR #108, opened 2022-06-17
language_php: Improve subsyntax highlighting

KatrinaKitten:
- SQL string marker keywords are now case insensitive
- SQL strings are recognized even with leading whitespace/parens
- Adjusted `<script>`/`<style>` tag detection to not break when attributes like `defer` are added
  - The same change has been PR'd to `language_html` (lite-xl/lite-xl#1043)
- Added `@` (error suppression) to operator list

### DIFF
--- a/plugins/language_php.lua
+++ b/plugins/language_php.lua
@@ -48,7 +48,7 @@ syntax.add {
     { pattern = "-?%.?%d+",                   type = "number"   },
     { pattern = "[%.%+%-=/%*%^%%<>!~|&%?:@]", type = "operator" },
      -- Variables
-    { pattern = "%$[%w_]+",                   type = "symbol" },
+    { pattern = "%$[%w_]+",                   type = "keyword2" },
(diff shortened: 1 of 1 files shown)


jgmdev:
Nice changes, could you also apply them (open another PR) for the 2.1 plugins branch? I adapted your previous PR with SQL strings support and added some additional changes to make it toggle-able and also settings gui support as seen here:

language_php

So adapting them I think will require more than a simple merge to 2.1 branch.

KatrinaKitten:
I may look into it at some point in the future, but life is crazy so I can't guarantee when I'll be available to do so, and I have no experience with the 2.1 changes yet. If you or someone else would like to pick up development on these changes for the 2.1 branch, please feel free.

jgmdev:
Applied the changes from this PR into master which code is now tailored for lite-xl 2.1.0, closing this PR as it targets older lite-xl 2.0 codebase. Thanks for the improvements!

PR #114, opened 2022-06-29
Add `primary_selection`

Guldoman:
This plugin uses `xclip` (or a similar command by changing the configuration) to implement middle mouse copy/paste.


jgmdev:
Seems that original requester didn't find any issues with this plugin so merging it :)

PR #118, opened 2022-07-17
Add syntax highlighting to JSX tags

dheisom:
It only adds the highlighting to JSX tags using the function colors

Before:
JSX file without tags highlighting

After:
JSX file with tags highlighting


jgmdev:
Seems good, merging....

PR #120, opened 2022-07-25
rainbowparens.lua: Added an option for number of parens

AqilCont:
Rainbowparens hardcoded 5 parenthesis colors for the colors it cycles through, and I added an option to change this number.


jgmdev:
Looks good so merging

PR #124, opened 2022-09-09
Syntax highlighter for the MiniScript programming language

ThaCuber:
it's what the title says lol. I didn't differenciate between functions and normal identifiers because in MiniScript you can call a function without parentheses (in some cases which I won't go onto detail about them here), which makes ae and ae() the same thing (you can get a reference to a function using @ though). I did however manage to highlight some built-in things and all intrinsic functions. I had to use a Lua script for that because I felt like it would've been long unnecessary work if I did it manually.


ThaCuber:
to be clear, this is referring to the [MiniScript](https://miniscript.org/) programming language developed by Joe Strout, not any other bitcoin related language :eyes:

jgmdev:
Looks good so merging

ThaCuber:
epic

PR #129, opened 2022-10-13
Add language file for .erb files

sathishmanohar:
erb files typically has filename.html.erb extension they contain html markup and embedded ruby code this file should properly syntax highlight both html and ruby parts


adamharrison:
So, just to keep a record, as talked about on discord: please retarget this for the 2.1 branch.

PR #130, opened 2022-10-14
Add language plugin for ERB files

sathishmanohar:
erb files typically has filename.html.erb extension they contain html markup and embedded ruby code this file should properly syntax highlight both html and ruby parts

### DIFF
--- a/plugins/language_erb.lua
+++ b/plugins/language_erb.lua
@@ -1,0 +1,55 @@
+-- mod-version:3
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "html-eruby",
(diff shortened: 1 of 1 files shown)


plugins/language_erb.lua
```diff
@@ -0,0 +1,55 @@
+-- mod-version:3
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "html-eruby",
+  files = { "%.htm%.erb$", "%.html%.erb$", "%.erb$" },
```
Guldoman, line 6:
```suggestion
  files = { "%.html?%.erb$", "%.erb$" },
```

sathishmanohar: Make the change. Will make the same in my other PR for 2.0.5 version as well.

PR #134, opened 2022-10-18
Minimap Rework with Scrollbar

adamharrison:
Reworked minimap to use Guldoman's new scrollbar class.

### DIFF
--- a/plugins/minimap.lua
+++ b/plugins/minimap.lua
@@ -398,6 +398,7 @@ function MiniMap:on_mouse_pressed(button, x, y, clicks)
   if y >= ty and y < ty + th then self.drag_start_offset = (y - ty) - th / 2 return self.percent end
   self.drag_start_offset = 0
   self.hovering.thumb = x >= tx and x < tx + tw and y >= ty and y < ty + th
+  self.dragging = self.hovering.thumb
   local lh = self.dv:get_line_height()
   percent = math.max(0.0, math.min((y - self.dv.position.y) / h, 1.0))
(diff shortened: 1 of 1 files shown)


adamharrison:
For clarity: this is essentially complete, with the exception of dragging the viewing window down the minimap. It's ever so slightly off.

Guldoman:
The minimap doesn't seem to cover the entire side of the view when the file is small. Is this intended?
image

IIRC the old plugin covered the full height.

adamharrison:
Oh, interesting. I thought that it didn't, specifically. I'll change that.

Guldoman:
It's working great!

Few minor things I noticed:
* Clicking on the thumb shouldn't scroll.
* Dragging from inside the thumb should keep the click location offset. For example if I start dragging from the top of the thumb, if I move the mouse it should still be at the top of the thumb.
* Clicking on the track shouldn't highlight the thumb.
* Moving the mouse after clicking on the track should either not do anything or make the thumb move immediately under the mouse.

adamharrison:
> Clicking on the thumb shouldn't scroll.
> Dragging from inside the thumb should keep the click location offset. For example if I start dragging from the top of the thumb, if I move the mouse it should still be at the top of the thumb.

Yeah; this is the same problem, due to it not taking into account offset.

> Clicking on the track shouldn't highlight the thumb.

This actually seems to happen on the normal scrollbar. Do we want to remove it specifically for minimap?

> Moving the mouse after clicking on the track should either not do anything or make the thumb move immediately under the mouse.

It currently does nothing. I think that's fine?

adamharrison:
OK, fixed account offset.

PR #138, opened 2022-10-28
Add Case Transformer plugin

merlindiavova:
A simple plugin to transform strings to and from PascalCase, camelCase, snake_case, SCREAMING_SNAKE_CASE, dot.case, kebab-case, backslash\case, slash/case etc.

This is my first attempt at writing Lua, so a review is probably necessary especially around the string matching.

Hopefully this is useful to people

### DIFF
--- a/plugins/strtransform.lua
+++ b/plugins/strtransform.lua
@@ -1,0 +1,288 @@
+-- mod-version:3
+local core = require "core"
+local command = require "core.command"
+
+local function is_lower(str)
+  local is_lowercase = true
+  for c in string.gmatch(str, ".") do
+      if not string.match(c,"%l")
+          and not string.match(c,"%d")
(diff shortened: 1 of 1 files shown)


adamharrison:
Hey! Sorry for the delay on this. I'm doing a big clean up of PRs, and just got to this one. Looks great! Tried a few, works totally fine.

One thing, though; it crashes lite for me if you don't have a selection (presumably it tries to transform the entire document).

Can you add predicate for your commands that makes it only function if there's a selection? Seems to me you'd never really want to transform the entire document.

adamharrison:
Also, can you add the plugin to the `manifest.json`?

merlindiavova:
Sure thing, sorry been super busy. Will get to this shortly

PR #142, opened 2022-11-03
Removed extra item in the plugins list

techie-guy:


Guldoman:
@takase1121 should we add a duplicate checker to the script?

adamharrison:
@Guldoman If we're gonna merge #140 we should probably just generate it from the manifest.

Guldoman:
Yeah, makes sense.

PR #145, opened 2022-11-04
Added extra file extensions to language_sh.lua

techie-guy:


plugins/language_sh.lua
```diff
@@ -3,7 +3,7 @@ local syntax = require "core.syntax"
 
 syntax.add {
   name = "Shell script",
-  files = { "%.sh$" },
+  files = { "%.sh$", "%.bash$", ".bashrc", ".bash_profile", ".profile" },
```
jgmdev, line 6: You forgot to escape the `.` with `%,` also adding `^` at beggining to only match files that start as written and `$` end also as written:

```lua
"^%.bashrc$", "^%.bash_profile$"
```

techie-guy: Oh, I'm sorry. Let me fix that real quick

techie-guy: fixed it :)

PR #147, opened 2022-11-06
Add custom caret plugin

techie-guy:
Plugin to customize the caret in the editor
Thanks to @Guldoman for the initial example code

Features
   1. Change the Color and Opacity of the caret
   2. Change the Shape of the caret
   3. Change the dimensions of the caret

More Details in the ```plugins/custom_caret.lua``` file


jgmdev:
To prevent unnecessary commits on your Pull Requests you always need to pull from the branch where you are going to submit a PR before submitting your changes. It will prevent this:

unneeded-commits

For this to properly work you should configure your git to `rebase` instead of `merge` when pulling to always put your local changes on top of current git branch changes. You have two options to achieve this:

**Only rebase pulls for current repository (you need to `cd` into repo dir):**
`git config pull.rebase true`

**Rebase pulls for all git repositories (can be run from anywhere):**
`git config --global pull.rebase true

Then you can safely run `git pull` and if you have any committed changes they will be rebased on top of latest commits when pulling from the working branch.

plugins/custom_caret.lua
```diff
@@ -0,0 +1,46 @@
...
+  elseif (config.plugins.custom_caret.shape == "Underline") then
+    config.plugins.custom_caret.width = caret_top_width
+    config.plugins.custom_caret.height = style.caret_width*1.5
+    y = y+lh
+  else
+    config.plugins.custom_caret.width = style.caret_width
+    config.plugins.custom_caret.height = lh
+  end
+
+  renderer.draw_rect(x, y, config.plugins.custom_caret.width, config.plugins.custom_caret.height, config.plugins.custom_caret.color)
```
jgmdev, line 45: Noticed that you also aren't using a default caret color in case the user didn't set one so you would need to use `style.caret` as default by adding it to the config table of the plugin.

plugins/custom_caret.lua
```diff
@@ -0,0 +1,46 @@
...
+local core = require "core"
+local style = require "core.style"
+local config = require "core.config"
+local DocView = require "core.docview"
+
+config.plugins.custom_caret = {}
+
+function DocView:draw_caret(x, y)
+  local lh = self:get_line_height()
+  local caret_top_width = math.ceil(self:get_font():get_width("a"))
```
jgmdev, line 28: as with `config.plugins.custom_caret.height` you can drop `config.plugins.custom_caret.width` and use `caret_top_width` directly which should be maybe renamed to `caret_width`.

techie-guy: but I use ```config.plugins.custom_caret.height``` and ```config.plugins.custom_caret.width```, so that they can be changed by the user, if they think the caret is too thick/thin.

techie-guy:
Thanks, I will keep that in mind, next time I create another pull request.

plugins/custom_caret.lua
```diff
@@ -0,0 +1,50 @@
...
+    config.plugins.custom_caret.width = caret_top_width
+    config.plugins.custom_caret.height = self:get_line_height()
+  elseif (current_caret_shape == "underline") then
+    config.plugins.custom_caret.width = caret_top_width
+    config.plugins.custom_caret.height = style.caret_width*1.5
+    y = y+self:get_line_height()
+  else
+    config.plugins.custom_caret.width = style.caret_width
+    config.plugins.custom_caret.height = self:get_line_height()
+  end
```
Guldoman, line 47: If you set the value to `config.plugins.custom_caret.*` here, it will overwrite what the user has set.

techie-guy: oh yea thanks, I think I need some more work on the plugin, I will mark this pull request as a draft, and notify when it's ready:)

techie-guy:
closed this pull request, will make a better plugin
sorry

PR #152, opened 2022-11-14
language_rust: better lifetime syntax support

0x42656E:
This PR will add following pattern (line 21):
```lua 
{ pattern = "[%a_][%w_]*()<%f['%a]",  type = {"function", "normal" } 
``` 

before
after

improvements are welcome :slightly_smiling_face:


jgmdev:
I'm not familiar with rust, so question: Does <'alpha only occurs on function calls?

0x42656E:
On structs too. Thanks for making me think of this.

I would solve this as follows:

```lua
  { pattern = "[%a_][%w_]*()<()'[%a]*()>()%(", type = {"function", "normal", "keyword2", "normal", "normal" } },
```
If there is a better method, let me know.

jgmdev:
You would need to use a regex with look ahead declaration `(?=[pattern])` instead of a lua pattern. Lua patterns do also have a look ahead, the one you used `%f` but they are more limited so you would need to do something like:

```lua
{ regex = "[:alpha:][\\w\\s]*(?=<[\\w\\d,']+>\\s*\\()", type = { "function" } },
```

Haven't tested that regex but just to give you an idea.

0x42656E:
Sadly Lite-XL crashes when i use your regex.
I had this also several times when using regex in Lite-XL for syntax matching therefore would like to avoid using regex in Lite-XL 

It is a bit dirty but here would be my final solution:

```lua
{ pattern = "fn[%s]+()[%a_][%w_]*%f[(<]", type = {"keyword", "function" } }
```

jgmdev:
Tested the regex here https://www.regextester.com/ and it worked after placing `[:alpha:]` in another pair of square brackets -> `[[:alpha:]]`

jgmdev:
After getting back to my machine tested and this version works perfectly:

```lua
{ regex = "[[:alpha:]][\\w]*(?=\\s*<[\\w\\s,']+>\\s*\\()",
  type = "function"
},
```
rust-fn-template

0x42656E:
Good job,  then i will push your version.

PR #157, opened 2022-11-21
feat: add group method for fontconfig plugin

TorchedSammy:
adds a method to load a range of fonts as a group and allows `group` in the use function


TorchedSammy:
(and also does mod version 3)

PR #164, opened 2022-12-06
gitstatus: color icons as well

cisoun:
Hi.

I propose the following change for the the `gitstatus` module:
- colorize the icons in the TreeView according to the Git status of the files.

Preview:
Capture d’écran 2022-12-06 à 23 32 26


jgmdev:
Can you make it toggleable and also add a config_spec for settings ui? Since some may prefer icon colors provided by plugins like nonicons.

cisoun:
Hello, I've added an option to toggle the colored icons: `config.plugins.gitstatus.color_icons = true`

Guldoman:
Thank you! Would you mind also bumping the plugin version in `manifest.json`?

cisoun:
Hey @Guldoman! Bumped to 0.2.

plugins/gitstatus.lua
```diff
@@ -39,6 +47,17 @@ function TreeView:get_item_text(item, active, hovered)
...
+  -- Override TreeView's get_item_icon to add modification color
+  local treeview_get_item_icon = TreeView.get_item_icon
+  function TreeView:get_item_icon(item, active, hovered)
+    local character, font, color = treeview_get_item_icon(self, item, active, hovered)
+    if cached_color_for_item[item.abs_filename] then
+      color = cached_color_for_item[item.abs_filename]
+    end
+    return character, font, color
+  end
+end
```
Guldoman, line 60: What do you think about something like:
```suggestion
-- Override TreeView's get_item_icon to add modification color
local treeview_get_item_icon = TreeView.get_item_icon
function TreeView:get_item_icon(item, active, hovered)
  local character, font, color = treeview_get_item_icon(self, item, active, hovered)
  if config.plugins.gitstatus and config.plugins.gitstatus.color_icons and cached_color_for_item[item.abs_filename] then
    color = cached_color_for_item[item.abs_filename]
  end
  return character, font, color
end
```

This way changing the option at runtime will apply immediately without a restart.

cisoun: Although I like this approach, the first solution would override `TreeView:get_item_icon` once and would not require checking also `config.plugins.gitstatus` and `config.plugins.gitstatus.color_icons` each 5 seconds for each file, this would add a bit overhead don't you think?

Guldoman: The overhead is 0 (even considering that this function is called at every redraw, not only every 5 seconds, for each visible entry).
This is basically the standard way plugins override things. We would be in big trouble if this had a big overhead...

cisoun: Alright. It's still some additional checks to do but we should be fine for now. Let's go with your solution then.

PR #167, opened 2022-12-09
display selection highlights in/on the scrollbar

KELs7:


adamharrison:
Looks good; does anyone have any opinions on merging this into selectionhiglight?

I have no objection; but one good thing to do, could be to include config options for this plugin, so you can toggle the scrollbar highlight on/off if people don't want that feature.

You feel up to adding that in, @KELs7 ? Configs look like this (pulled from `centerdoc`):

```
local config = require "core.config"
config.plugins.centerdoc = common.merge({
  enabled = true,
  zen_mode = false
}, config.plugins.centerdoc)
```

adamharrison:
Oh, also, try to avoid hardcoding colors; if you're gonna use a color, try and pick one that makes sense from `style`, or, if you need a new one, definitely let it be a config variable.

plugins/selectionhighlight.lua
```diff
@@ -56,3 +56,43 @@ function DocView:draw_line_body(line, x, y)
...
+  local scrollbar = self.v_scrollbar
+
+  local mw = 5  --marker width
+  local mh = 8  -- marker height
+  local cmh = 4  --current selection marker height
+  local cm_color =  {common.color "#DA70D6"} --current selection marker color
+  
+  local line1, col1, line2, col2 = get_selection(self.doc, true)
+  
+  if line1 == line2 and col1 ~= col2 then
```
adamharrison, line 77: So, this will only be drawn if we're selecting exactly one line? It won't show anything for multiline selections? Is that correct?

plugins/selectionhighlight.lua
```diff
@@ -56,3 +56,43 @@ function DocView:draw_line_body(line, x, y)
...
+    local selection = self.doc:get_text(line1, col1, line2, col2)
+    
+    if self.doc and scrollbar then
+      local x,y,w,h = scrollbar:get_track_rect()
+      
+      for line, line_text in pairs(self.doc.lines) do
```
adamharrison, line 83: `self.doc.lines` is an array, so you'll want to use `ipairs` if you're going to iterate through it.

plugins/selectionhighlight.lua
adamharrison: I guess this matches the behaviour above, so that's OK.

plugins/selectionhighlight.lua
```diff
@@ -56,3 +56,43 @@ function DocView:draw_line_body(line, x, y)
...
+local old_draw_docview_scrollbar = DocView.draw_scrollbar
+
+function DocView:draw_scrollbar()
+  old_draw_docview_scrollbar(self)
+  local scrollbar = self.v_scrollbar
+
+  local mw = 5  --marker width
+  local mh = 8  -- marker height
+  local cmh = 4  --current selection marker height
+  local cm_color =  {common.color "#DA70D6"} --current selection marker color
```
Guldoman, line 73: You're missing a `local common = require "core.common"`.

Guldoman (commented):
The location of the marks in the scrollbar is not really precise relative to the scrollbar thumb.

The location of the marks is completely off with the `minimap` plugin.
IIRC the `minimap` plugin has a way to specify where to draw marks, so you could support that if the minimap is enabled.
Otherwise you could be strict and only show the marks when the scrollbar is the "original" one.

The marks don't expand with the scrollbar, is this intended?

KELs7:
> Looks generally good; does anyone have any opinions on merging this into selectionhiglight?
> 
> I have no objection; but one good thing to do, could be to include config options for this plugin, so you can toggle the scrollbar highlight on/off if people don't want that feature.
> 
> You feel up to adding that in, @KELs7 ? Configs look like this (pulled from `centerdoc`):
> 
> ```
> local config = require "core.config"
> config.plugins.centerdoc = common.merge({
>   enabled = true,
>   zen_mode = false
> }, config.plugins.centerdoc)
> ```



> 

ok. will look into this and the other suggestions made. thanks

plugins/selectionhighlight.lua
KELs7: oh sorry. will add "common".  it's added in mine locally.

yes marks don't expand with the scrollbar.

PR #171, opened 2022-12-18
Improved language_wren.lua

ThaCuber:
the syntax highlighter *was* fine until I saw it behaving more like Lua's, as if someone just copied Lua's syntax highlighter and pasted it here, then modified it a bit... so I improved the thing
no class or variable outside a class can start with `_`, so I removed the `_` out of `[%a_]` and just left `%a`
and there cannot be a hanging `.` in front or behind a number either, there can't be at the front since `1.floor` and `1.atan(2)` are valid syntax, and there can't be behind because idk
also `if`, `while` and `for` need parentheses and the function section was highlighting those as functions when they aren't
none of the commits make sense either so don't mind them lmao


ThaCuber:
\*dies*
image

ThaCuber:
ok that should do it 👍

ThaCuber:
that commit name is unclear as fuck, to see those changes you gotta check the commit description

jgmdev:
it is ready for merging?

ThaCuber:
I wanted to highlight the usage of `Fn.new` too so people know it's a standalone function, but I'd say it's done without that

jgmdev:
> I wanted to highlight the usage of `Fn.new` too so people know it's a standalone function, but I'd say it's done without that

"Fn()%.()new", {"type", "type", "type"}?

ThaCuber:
nah, more like the whole `Fn.new`, highlighted as a "keyword2" but I don't know if that is even possible without breaking something else, nor if it will look good on all themes

jgmdev:
so should I merge it as is?

ThaCuber:
yeah, go ahead, it's not that necessary anyways

unless you're me right now
image

ThaCuber:
I have completely forgotten that Wren also has [annotations](https://wren.io/classes#attributes) so I still needed to add more 😔

jgmdev:
another PR :)

PR #173, opened 2022-12-24
Add `tab_switcher`

Guldoman:
This plugin adds a couple of commands to allow switching between tabs by searching their name.


Titousensei:
Very useful! This would be very nice as a right-click popup on the tab bar.

jgmdev:
What do you think about binding this to `alt+p` by default?

Guldoman:
> What do you think about binding this to `alt+p` by default?

Are there other editors that use that shortcut?

jgmdev:
Since we are already using ctrl+p, ctrl+shift+p I thought alt+p would be convenient

Guldoman:
Should `alt+p` be `tab-switcher:tab-list` or `tab-switcher:tab-list-current-split`?

jgmdev:
`tab-switcher:tab-list` and maybe alt+shift+p `tab-switcher:tab-list-current-split`

PR #175, opened 2022-12-26
Added annotations to language_wren.lua

ThaCuber:
forgot to put them in pull request #171 so I now put them here lol
there are still some things that I can't solve because of how lua pattern matching works, like something like `#!a = 1 3  5 25 235 235` and `#!a =1)` get completely highlighted but `#!a = 1 ()` doesn't highlight neither of the parentheses


Guldoman:
You can use regexes if you prefer. Simply replace `pattern` with `regex`.
For complex regexes that would require a lot of string escapes I suggest using Lua's raw strings:
```lua
[[this is a raw string\n<- this is not a newline, but \ followed by n

^ this is a newline]]
```

ThaCuber:
I can do that???

Guldoman:
I mean, it's documented on the syntax highlighting guide in the website, but to be fair it's literally a single phrase and not expanded further.

ThaCuber:
couldn't fix annotations but did however made `Fn.new` highlight as a literal

PR #176, opened 2022-12-26
Add smartopenselected plugin

Titousensei:
New plugin to open the selection as a filename or path in the project. Combination of openselected and ctrl+p.
Path separators are used to replace ".", so import path can also be found.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1121,6 +1121,13 @@
       "mod_version": "3"
     },
     {
+      "description": "Opens the selected filename or path in project. Useful to open imports.",
+      "version": "0.1",
+      "path": "plugins/smartopenselected.lua",
+      "id": "smartopenselected",
+      "mod_version": "3"
+    },
+    {
       "description": "Smooth caret animation *([gif](https://user-images.githubusercontent.com/20792268/139006049-a0ba6559-88cb-49a7-8077-4822445b4a1f.gif))*",
       "version": "0.1",
       "path": "plugins/smoothcaret.lua",
--- a/plugins/smartopenselected.lua
+++ b/plugins/smartopenselected.lua
@@ -1,0 +1,48 @@
+-- mod-version:3
+local core = require "core"
+local command = require "core.command"
+local keymap = require "core.keymap"
+local common = require "core.common"
+local contextmenu = require "plugins.contextmenu"
+
+
+command.add("core.docview!", {
+  ["smart-open-selected:smart-open-selected"] = function(dv)
+    local doc = dv.doc
+    if not doc:has_selection() then
+      core.error("No text selected")
+      return
+    end
+
+    local text_orig = doc:get_text(doc:get_selection())
+    text_orig = text_orig:match( "^%s*(.-)%s*$" )
+    
+    -- transform java/python imports to paths
+    local text_path, num = text_orig:gsub("[.]", PATHSEP)
+    
+    -- keep the last . in case the path contains a file extension
+    local text_keep_extension, num = text_orig:gsub("[.]", PATHSEP, num - 1)
+
+    -- trim whitespace from the ends
+
+    for dir, item in core.get_project_files() do
+      if item.type == "file" and (
+         string.find(item.filename, text_orig)
+         or string.find(item.filename, text_path)
+         or string.find(item.filename, text_keep_extension)
+      ) then
+        local path = (dir == core.project_dir and "" or dir .. PATHSEP)
+        local filepath = common.home_encode(path .. item.filename)
+        core.root_view:open_doc(core.open_doc(common.home_expand(filepath)))
+      end
+    end
+  end,
+})
+
+
+contextmenu:register("core.docview", {
+  { text = "Smart Open Selection",  command = "smart-open-selected:smart-open-selected" }
+})
+
+
+keymap.add { ["ctrl+shift+alt+p"] = "smart-open-selected:smart-open-selected" }


adamharrison:
Hey @Titousensei, sorry about the delay; I'm going through now and merging PRs that need it; this looks fine. Can you just adjust it to use the new manifest system? Once that's done, I'll merge.

plugins/smartopenselected.lua
```diff
@@ -0,0 +1,38 @@
+-- mod-version:3
+local core = require "core"
+local command = require "core.command"
+local keymap = require "core.keymap"
+local common = require "core.common"
+local contextmenu = require "plugins.contextmenu"
+
+
+command.add("core.docview!", {
```
Guldoman, line 9: Might be a good idea to check if a selection exists in the command predicate.

plugins/smartopenselected.lua
```diff
@@ -0,0 +1,38 @@
...
+        local path = (dir == core.project_dir and "" or dir .. PATHSEP)
+        local filepath = common.home_encode(path .. item.filename)
+        core.root_view:open_doc(core.open_doc(common.home_expand(filepath)))
+      end
+    end
+  end,
+})
+
+
+contextmenu:register("core.docview", {
```
Guldoman, line 33: Same as above.

README.md
```diff
@@ -194,6 +194,7 @@ but only with a `url` must provide a `checksum` that matches the existing plugin
...
 | [`smallclock`](plugins/smallclock.lua?raw=1)                                                                   | Displays the current time in the corner of the status view                                                                                                                                                                                                                                       |
+| [`smartopenselected`](plugins/smartopenselected.lua?raw=1)                                                     | Opens the selected filename or path in project. Useful to open imports.                                                                                                                                                                                                                          |
```
Guldoman, line 197: You can remove this change as it will be automatically generated from the manifest.

manifest.json
```diff
@@ -1105,6 +1105,13 @@
       "id": "smallclock",
       "mod_version": "3"
     },
+    {
+      "description": "Opens the selected filename or path in project. Useful to open imports.",
+      "version": "0.1",
+      "path": "plugins/smartopenselected.lua",
+      "id": "smartopenselected",
+      "mod_version": "3"
+    },
```
Guldoman, line 1114: If you want to, you can add a `name` parameter for a "normal", user-visible name.

plugins/smartopenselected.lua
Titousensei: Default behavior is an error message in the status bar "no text selected", which seems appropriate.

Titousensei:
Updated with manifest

adamharrison:
OK, reviewing now.

adamharrison:
All right, so I'm looking at this, I'm not quite understanding why you're replacing `.` with a path separator.

Like, for example, say I'm looking at the plugins repo as my project.

If I have the string `manifest.json` selected, this *does not* open manifest.json, because it's replacing it to be `manifest/json`, but not doing the same when it's checking through the iterations of project files.

Also, `ctrl+shift+alt+p` doesn't actually work for me; because we don't normalize the stroke order (we'll fix this), but for now, change it to `ctrl+alt+shift+p` so that it functions correctly.

Titousensei:
> All right, so I'm looking at this, I'm not quite understanding why you're replacing `.` with a path separator.

This is for python and java imports, for example.

> If I have the string `manifest.json` selected, this _does not_ open manifest.json, because it's replacing it to be `manifest/json`, but not doing the same when it's checking through the iterations of project files.

I see. I did not think of that. I'm not sure I can handle both cases.
 
> Also, `ctrl+shift+alt+p` doesn't actually work for me; because we don't normalize the stroke order (we'll fix this), but for now, change it to `ctrl+alt+shift+p` so that it functions correctly.

ok.

adamharrison:
> This is for python and java imports, for example.

Gotcha.

> I see. I did not think of that. I'm not sure I can handle both cases.

What I'd do, is just keep two copies of the string. Check for the presence of the transformed one `or` check for the presence of the unmodified one. That's easy enough; and would probably work.

Titousensei:
> Also, ctrl+shift+alt+p doesn't actually work for me; because we don't normalize the stroke order (we'll fix this), but for now, change it to ctrl+alt+shift+p so that it functions correctly.

`ctrl+shift+alt+p` works for me (linux), but `ctrl+alt+shift+p` does not work.

Titousensei:
Updated to try with PATHSEP, or original string, or PATHSEP except for the last `.` in case of extensions.

adamharrison:
> ctrl+shift+alt+p works for me (linux), but ctrl+alt+shift+p does not work.

Ok, well, whatever; we'll just fix this on the primary application, so not a huge deal wither way.

> Updated to try with PATHSEP, or original string, or PATHSEP except for the last . in case of extensions.

So, looks like you added some comments; but lua uses `--` for comments, not `#`, so this is now throwing a syntax error. Can you swap for `--`?

Titousensei:
> So, looks like you added some comments; but lua uses `--` for comments, not `#`, so this is now throwing a syntax error. Can you swap for `--`?

Argh! Fixed.

adamharrison:
All right! Looks good. Merging.

One thing to note, that might be interesting; in the absence of a selection, maybe just get where the cursor is, and grab everything around it that isn't a word boundary or a `.` or `/`. 

This would let people do things like bind a ctrl+alt+lclick or something, so that you'd be able to just click on an identifier and go there.

But it's not at all necessary for this to be useful; so I'll just merge as-is now.

PR #177, opened 2022-12-26
Extend eval plugin with eval:selected

Titousensei:
- if text is selected, replace this text with eval result
- if no text is selected, eval the current line and insert the result in the next line with "= " prefix.

Adding `ctrl+alt+return` default keymap

### DIFF
--- a/plugins/eval.lua
+++ b/plugins/eval.lua
@@ -1,6 +1,8 @@
 -- mod-version:3
 local core = require "core"
 local command = require "core.command"
+local contextmenu = require "plugins.contextmenu"
+local keymap = require "core.keymap"
 
 
 local function eval(str)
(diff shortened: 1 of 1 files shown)


plugins/eval.lua
```diff
@@ -29,4 +31,23 @@ command.add("core.docview", {
...
+    else
+      local line = dv.doc:get_selection()
+      local text = dv.doc.lines[line]
+      dv.doc:insert(line+1, 0, "= " .. eval(text) .. "\n")
+    end
+  end,
 })
+
+
+contextmenu:register("core.docview", {
```
jgmdev, line 48: Not sure if the predicate should be turned to a function that besides checking if current view is a `DocView` also checks that the current doc file extension ends in `.lua`, since we can't eval anything besides lua code and math expressions compatible with lua.

But if I recall correctly context menu ignores the given predicate and uses that of the assigned command, so this predicate could even be nil and it would use the one for the command above.

jgmdev (approved):

plugins/eval.lua
Titousensei: Because of the math expressions, which very similar in many languages (and even useful in .txt files sometimes), I don't think we should restrict on the doc file extension anyway.

plugins/eval.lua
```diff
@@ -29,4 +31,23 @@ command.add("core.docview", {
...
+  { text = "Evaluate Selected",  command = "eval:selected" }
+})
+
+
+keymap.add { ["ctrl+return"] = "eval:selected" }
```
jgmdev, line 53: I was going to merge this but noticed that `ctrl+return` is already binded in core to insert empty line after current line

PR #178, opened 2022-12-26
Add new pluging git:open

Titousensei:
Given a git commit (default HEAD), open the files that are in the project and part of this commit.


plugins/gitopen.lua
```diff
@@ -0,0 +1,43 @@
...
+local command = require "core.command"
+local common = require "core.common"
+
+
+local function cmd(cmd)
+  local fp = io.popen(cmd, "r")
+  local res = fp:read("*a")
+  fp:close()
+  return res
+end
```
Guldoman, line 12: I haven't tested the plugin yet, but I'd suggest using our `process` API instead of relying on `io.popen`.
The main reason is that the `process` API allows async operations, so that the UI doesn't block.

You can find some example of `process` usage in a few plugins in this repo (`gitstatus`).
The API documentation is at https://github.com/lite-xl/lite-xl/blob/c29b1c2cb9b50f11cfe3b3bd6a9d59ff0fd8488d/docs/api/process.lua.

Titousensei:
Rewrote the exec command to use `process` in coroutine as suggested.

plugins/gitopen.lua
```diff
@@ -0,0 +1,53 @@
...
+
+  for dir, item in core.get_project_files() do
+    local key = dir .. PATHSEP .. item.filename
+    if git_files[key] then
+      core.root_view:open_doc(core.open_doc(item.filename))
+    end
+  end
+end
+
+command.add("core.docview", {
```
Guldoman, line 36: Wouldn't it be better to allow calling the command even if the user isn't in a `DocView`?

plugins/gitopen.lua
```diff
@@ -0,0 +1,53 @@
...
+  ["git:open"] = function(dv)
```
Guldoman, line 37: I would use a more descriptive name, like `gitopen:open-from-commit` or something along those lines.

Titousensei:
Updated to use from any context (not just docview), and renamed command to `gitopen:open-from-commit`, as suggested.

adamharrison:
@Titousensei ; same deal as with #176; just adjust to use manifest, and I'll review promptly.

Titousensei:
Updated with manifest

Titousensei:
Updated with manifest

adamharrison:
OK, reviewing now.

adamharrison:
OK; just one thing; on your `exec`, when you do read_stdout; that'll only read out a buffer of 4096 bytes maximum in normal circumstances.  Should work in most cases, but just a heads up.

We may actually fix this in the editor itself; as it doesn't really make sense to not have it just read everything it can, absent an argument. So I'm happy to merge this as is; works fine.

PR #179, opened 2022-12-26
`spellcheck`: Add cache

Guldoman:
This improves performance a lot in some cases (try the `README` of this repo).
I haven't tested this extensively.

We need a generic way to add a cache based on the `Highlighter`.

### DIFF
--- a/plugins/spellcheck.lua
+++ b/plugins/spellcheck.lua
@@ -5,6 +5,7 @@ local config = require "core.config"
 local command = require "core.command"
 local common = require "core.common"
 local DocView = require "core.docview"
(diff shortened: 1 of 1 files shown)


jgmdev:
I know this isn't related to current PR but the predicate of the context menu items should also be fixed to not show them if spellcheck isn't enabled or current doc doesn't matches the filters list.

Guldoman:
Mmm might still be useful as an on-demand thing.
Maybe we could set the predicate to return `false` if the plugin is disabled.

Guldoman:
Merging for now.
We'll manage the predicate changes in another PR.

Issue #174, opened 2022-12-24
plugin bug lfautoinsert


beastehg:
https://github.com/lite-xl/lite-xl-plugins/blob/master/plugins/lfautoinsert.lua
It even closes single tags and those that do not need
https://i.imgur.com/6GjFiwY.png
I found the line which is responsible for all this nonsense but I don't know regular expressions (could someone fix it)
https://i.imgur.com/60ip6bz.png

Titousensei:
Fix is merged. Please close this issue.

PR #180, opened 2022-12-27
Fix lfautoinsert

Titousensei:
Fix for issue #174

In html files, tags starting with `<!` or `</` should not trigger the plugin to create a closing tag. Other tags should.

Test:
- create empty file called `test.html`
- type: `<!DOCTYPE html>` and press Enter
- verify that the plugin is *not* creating a closing tag
- type `<meta>` and press Enter
- verify that the plugin *is* creating a closing tag `</meta>`


jgmdev:
Thanks for fixing this annoyance!

PR #187, opened 2023-01-01
exec: NEW command exec:replace-from-document

atorr0:
This PR barely improves the existing "exec:replace" command with a new option that pipes the document content to the STDIN of the CLI command.

An example of the execution could be calling the `sort` command.

### DIFF
--- a/plugins/exec.lua
+++ b/plugins/exec.lua
@@ -1,6 +1,7 @@
 -- mod-version:3
 local core = require "core"
 local command = require "core.command"
+local DocView = require "core.docview"
 
 
 local function exec(cmd, keep_newline)
@@ -8,6 +9,30 @@ local function exec(cmd, keep_newline)
   local res = fp:read("*a")
   fp:close()
   return keep_newline and res or res:gsub("%\n$", "")
+end
+
+
+local function get_active_view()
+  if getmetatable(core.active_view) == DocView then
+    return core.active_view
+  end
+  return nil
+end
+
+
+local function execw(cmd, v, keep_newline)
+  local fp = io.popen(cmd, "w")
+  fp:write(tostring(v))
+  fp:close()
+end
+
+
+local function exec_doc(cmd, keep_newline)
+  local tmp = exec("mktemp -t lite-xl.tmp.XXXXXXXXXX", false)
+  local av_doc = get_active_view().doc
+  execw(cmd .. " 2>&1 >" .. tmp, av_doc, keep_newline)
+
+  return exec("cat " .. tmp, keep_newline)
 end
 
 
@@ -40,6 +65,19 @@ command.add("core.docview", {
       submit = function(cmd)
         dv.doc:replace(function(str)
           return exec(
+            "printf %b " .. printfb_quote(str:gsub("%\n$", "") .. "\n") .. " | eval '' " .. shell_quote(cmd),
+            str:find("%\n$")
+          )
+        end)
+      end
+    })
+  end,
+
+  ["exec:replace-from-document"] = function(dv)
+    core.command_view:enter("Replace With Result Of Command From Piped Document Content", {
+      submit = function(cmd)
+        dv.doc:replace(function(str)
+          return exec_doc(
             "printf %b " .. printfb_quote(str:gsub("%\n$", "") .. "\n") .. " | eval '' " .. shell_quote(cmd),
             str:find("%\n$")
           )


plugins/exec.lua
```diff
@@ -11,6 +12,30 @@ local function exec(cmd, keep_newline)
 end
 
 
+local function get_active_view()
+  if getmetatable(core.active_view) == DocView then
+    return core.active_view
+  end
+  return nil
+end
```
Guldoman, line 20: No need for this function as you get the active `DocView` from the command predicate, so you can pass it directly to the other functions.

Guldoman (commented):
I wonder if it would be a good idea to use the `Process` API instead of `io.popen`, so that the editor doesn't lock up during the execution of the command.
With the Process API you also won't need to create temporary files to read/write from/to the process.

atorr0:
Thank you @Guldoman for your review.

I'm not fluent in Lua, so feel free to modify this PR.
About the Process API topic, I've tried to be conservative on changes, so I have mainly followed the same approach of the plugin. Nevertheless I was not aware of the benefits of the [Process API](https://lite-xl.com/en/tutorials/overview/process) so I'll update it accordingly.

plugins/exec.lua
```diff
@@ -1,6 +1,8 @@
 -- mod-version:3
 local core = require "core"
 local command = require "core.command"
+local DocView = require "core.docview"
+local process = require "process"
```
Guldoman, line 5: No need to do this, `process` is a global value.

plugins/exec.lua
```diff
@@ -11,6 +11,19 @@ local function exec(cmd, keep_newline)
...
+  local proc = process.start { "sh", "-c", cmd }
+  proc:write(tostring(dv))
+  local res = ""
+  while true do
+    local rdbuf = proc:read_stdout()
+    if not rdbuf then break end
+    res = res .. rdbuf
+  end
+  return keep_newline and res or res:gsub("%\n$", "")
+end
```
Guldoman, line 24: For performance reasons I'd suggest
```suggestion
local function exec_rw(dv, cmd, keep_newline)
  local proc = process.start { "sh", "-c", cmd }
  proc:write(tostring(dv))
  local res = {}
  while true do
    local rdbuf = proc:read_stdout()
    if not rdbuf then break end
    if #rdbuf > 0 then table.insert(res, rdbuf) end
  end
  res = table.concat(res)
  return keep_newline and res or res:gsub("%\n$", "")
end
```

plugins/exec.lua
```diff
@@ -11,6 +11,19 @@ local function exec(cmd, keep_newline)
 end
 
 
+local function exec_rw(dv, cmd, keep_newline)
+  local proc = process.start { "sh", "-c", cmd }
+  proc:write(tostring(dv))
```
Guldoman, line 16: Does `tostring(dv)` do what you expect?

plugins/exec.lua
```diff
@@ -11,6 +11,19 @@ local function exec(cmd, keep_newline)
 end
 
 
+local function exec_rw(dv, cmd, keep_newline)
+  local proc = process.start { "sh", "-c", cmd }
```
Guldoman, line 15: Might be a good idea to check that this didn't fail. In that case a `core.error` message might be appropriate.

Guldoman (commented):
This still blocks the UI as you're looping on read, but that's probably fine I guess.

plugins/exec.lua
atorr0: It works as I expect: It transforms the content to a string (no extra newlines, nothing strange...)
Do you suggest a different approach?

plugins/exec.lua
atorr0: Done

plugins/exec.lua
atorr0: Finally, as you suspected, it didn't work. I've removed `dv` references and worked in a more direct approach.

atorr0:
Checked with `sort` command, but also with `for f in 2 3 4; do sleep 3; echo $f; done` commands and the editor is not blocked and the results are correct.

Side note: Thanks to dynamic `yieldTime`, not only the editor is not blocked but also it's responsive, quick to react for fast commands, and CPU usage is near 0% for longer commands.

plugins/exec.lua
```diff
@@ -11,6 +11,32 @@ local function exec(cmd, keep_newline)
 end
 
 
+local function exec_rw(cmd, keep_newline, str)
+  local proc = process.start { "sh", "-c", cmd }
```
Guldoman, line 15: I wonder if it would be more appropriate to either just execute the command without `sh`, or to add things like `cmd` for Windows.

plugins/exec.lua
```diff
@@ -11,6 +11,32 @@ local function exec(cmd, keep_newline)
...
+    if not rdbuf then break end
+    if #rdbuf > 0 then table.insert(res, rdbuf) end
+
+    coroutine.yield(yieldTime)
+    if yieldTime < 1 then
+      yieldTime = yieldTime * 2
+    end
+  end
+
+  if proc:returncode() == 127 then
```
Guldoman, line 31: I guess that this is this true just for `sh`.
If you decide to either make it execute the command directly, or execute the windows shell too, this might not be effective anymore.

plugins/exec.lua
atorr0: Thank you for your review.

If the shell is removed then all the features provided by a shell (like piping) will be lost, so IMHO the benefits of calling the shell totally outweights the costs.

About supporting Windows, AFAIK currently there're PowerShell and the "old" `cmd` (I don't know if ~command.com~ is still shipped), so a deeper investigation will be necessary.

plugins/exec.lua
atorr0: Sure, the original code of `exec.lua` was (and the other functions still are) bound to *nix / Bash, but if it's planned to boost this plugin to support Windows, then I think it could be done by using a pointer to a table with the needed functions to handle the underlying shells (Bash, PowerShell, cmd, etc.), initialized at load time.

Again, I'm not fluent in Lua so maybe this approach is not the best.

Guldoman: You could add a config option with the command to use. And maybe set the default one according to `PLATFORM`.

PR #189, opened 2023-01-01
add language_nix

Meerschwein:


README.md
```diff
@@ -123,6 +123,7 @@ but only with a `url` must provide a `checksum` that matches the existing plugin
...
 | [`language_nim`](plugins/language_nim.lua?raw=1)                                                               | Syntax for the [Nim](https://nim-lang.org) programming language                                                                                                                                                                                                                                  |
+| [`language_nix`](plugins/language_nix.lua?raw=1)                                                               | Syntax for the [Nix](https://nixos.wiki/wiki/Overview_of_the_Nix_Language) expression language                                                                                                                                                                                                   |
```
Guldoman, line 126: I think you can remove this, it'll be automatically generated.

plugins/language_nix.lua
```diff
@@ -0,0 +1,87 @@
...
+local string_interpolation = {
+  { pattern = {"%${", "}"}, type = "keyword2", syntax = {
+    patterns = default_patterns,
+    symbols = default_symbols,
+  }},
+  { pattern = "[%S][%w]*", type = "string" },
+}
+
+merge_tables(default_patterns, {
+  { pattern = "#.-\n",        type = "comment" },
```
Guldoman, line 39: I just noticed that many other language plugins use `.-\n`.
This is not needed, `.*` should be enough.

PR #194, opened 2023-01-08
keymap_export: add keymap_export

takase1121:
The script in `scripts/keymap-generator` no longer works.

Instead, we use Lite XL to run a plugin that exports the keymap.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -300,6 +300,13 @@
       "mod_version": "3"
     },
     {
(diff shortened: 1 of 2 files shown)


adamharrison:
Bit odd for it to not be a dictionary, no? But, either way, it works, and outputs JSON; so I'm good with it.

takase1121:
It was originally a dictionary, but I decided to sort the output so it was made into an array.

PR #195, opened 2023-01-09
Fix `autoinsert.lua`'s behavior on multicursor

AqilCont:
The previous behavior was that it tried to act on the first selection, essentially discarding the ongoing multicursor selection. Now, it properly ignores multicursor.


Guldoman (commented):
Might be a good idea to also limit this predicate:
https://github.com/lite-xl/lite-xl-plugins/blob/cc9a7ba354c28fb3397c3ab65cb83647be02a14f/plugins/autoinsert.lua#L99-L102
so that the commands don't get executed in the multi-cursor case too.

Guldoman (approved):
Just a couple of nits.

Guldoman:
Oh yeah, might be a good idea to bump the plugin version too.

AqilCont:
Sorry, WDYM bump the plugin version?

PR #197, opened 2023-01-10
Remove workaround for sorted `Doc:get_selection` issue

Guldoman:
As 2.1.1 contains the fix needed, let's remove the workaround.


adamharrison:
Ok, tested, all works, looks good. Merging.

PR #198, opened 2023-01-12
Add Lite Debugger plugin

Jan200101:
modified from https://github.com/slembcke/debugger.lua

allows developers to interactively debug the lua VM Lite-XL runs on.
different to upstream it also allows for breakpoints to be set on exposed C functions.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -113,6 +113,13 @@
       "mod_version": "3"
     },
     {
+      "description": "Debug Lite-XL's Lua VM interactively",
+      "version": "0.1",
+      "path": "plugins/lite-debugger.lua",
+      "id": "lite-debugger",
+      "mod_version": "3"
+    },
+    {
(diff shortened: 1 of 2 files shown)


Guldoman:
I wonder if it would be practical to add a command that enables the hooks (and disables them) so that it can be "active" without hurting performance even when not needed.
Otherwise I'd add a big disclaimer in the description that having this enabled *will* hurt performance.

adamharrison:
I think it's reasonable as-is; it does what it says, and people can simply disable it using the normal plugin disable mechanisms or uninstall it if they don't want. I figure if we get complaints about people saying they're getting slowdowns, and don't know why, and they have this installed, then yeah; let's mandate a change to the description, but for now, it's good enough, and works nicely.

adamharrison:
But in the interests of just getting things closed, I'll merge this now, and I'll modify the description to mention the terminal requirement.

adamharrison:
OK, we literally had our first casualty of someone having this problem, so I am going to modify the description and add in the warning lol.

PR #199, opened 2023-01-16
Add `svg_screenshot`

Guldoman:
This needs a few config options (like don't embed fonts, generate an HTML page, ...).

Might be worth it to investigate how to make animations with this.

The `base64` library this uses should probably be a different addon (even a native one maybe?).

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1110,6 +1110,13 @@
       "mod_version": "3"
     },
     {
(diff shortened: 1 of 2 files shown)


adamharrison:
I have some issues doing ctrl+f12, but seems to be my window manager. Rebound the keys, and works great. Super awesome plugin. 

Only thing is, it seems like my local image viewer doesn't see it correctly so the standards compliance may be wonky; but whatever, this is primarily for use on websites anyway. Merging.

PR #202, opened 2023-01-20
Add Nerdicons Plugin

techie-guy:
# Nerdicons Plugin

### **This is an extension/modification of the [nonicons plugins](https://github.com/lite-xl/lite-xl-plugins/blob/master/plugins/nonicons.lua)**

**Any icon can be searched by it's hex from [nerdfonts.com/cheat-sheet](https://nerdfonts.com/cheat-sheet)**

## How to use:
1. Download a nerd font that you like from [nerdfonts.com](https://nerdfonts.com)
2. Unzip the zip file that you downloaded.
3. Choose a font file from the unzipped directory and install it.
4. Copy the font file to the lite-xl data directory and rename it to "icon-nerd-font.ttf".
5. Restart lite-xl, the icons should appear.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -966,7 +966,7 @@
       "path": "plugins/nerdicons.lua",
       "id": "nerdicons",
       "mod_version": "3"
-    }
+    },
     {
       "description": "File icons set for TreeView. Download [font](https://github.com/yamatsum/nonicons/raw/6a2faf4fbdfbe353c5ae6a496740ac4bfb6d0e74/dist/nonicons.ttf) to your config/fonts folder",
       "version": "0.1",
--- a/manifest.json
+++ b/manifest.json
@@ -960,6 +960,14 @@
       "id": "net"
     },
     {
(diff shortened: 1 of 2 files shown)


techie-guy:
I have updated the manifest.json.
Is there anything else, I have to do?

techie-guy:
Okay:)

adamharrison:
OK; tried it, looks good! Happy to merge as is.

One of the things that you may be interested in though, is the new `files` key that you can add in the manifest.

It may not make sense for this plugin specifically; as you're supposed to choose the iconset you want.

But one thing you can do, is specify a default font url from nerdicons, and flag it as optional, so that things'll basically just work out of the box when someone uses a plugin manager to install (the plugin manager will download the zip to the plugins folder, and then unzip it). 

Though you'd have to set up some extra logic to rename the font as it comes out of the zip; it could be a nice addition.

But I'll merge this now; feel free to tack it on as another PR if you wanna do it; otherwise, we're still good anyway.

techie-guy:
Okay, will do it if I have any free time.

Thanks😁

PR #204, opened 2023-01-22
Added in an if guard to regexreplaceplugin.

### LINKED ISSUE #60: regex preview plugin error
When use [regexreplacepreview](https://github.com/lite-xl/lite-xl-plugins/blob/master/plugins/regexreplacepreview.lua?raw=1)
and type \n (linebreak) anywhere in regex program crashes, displaying error
lite-xl-tegex-previev2
Content of file error.txt:
```Error: ...paraf\progs\lite-xl/data/plugins\regexreplacepreview.lua:50: attempt to get length of local 'old_text' (a nil value)
stack traceback:
	...paraf\progs\lite-xl/data/plugins\regexreplacepreview.lua:50: in function 'regex_replace_file'
	...paraf\progs\lite-xl/data/plugins\regexreplacepreview.lua:101: in function 'suggest'
	D:\pparaf\progs\lite-xl/data/core\commandview.lua:178: in function 'update_suggestions'
	D:\pparaf\progs\lite-xl/data/core\commandview.lua:200: in function 'update'
	D:\pparaf\progs\lite-xl/data/core\rootview.lua:516: in function 'update'
	D:\pparaf\progs\lite-xl/data/core\rootview.lua:523: in function 'update'
	D:\pparaf\progs\lite-xl/data/core\rootview.lua:524: in function 'update'
	D:\pparaf\progs\lite-xl/data/core\rootview.lua:524: in function 'update'
	D:\pparaf\progs\lite-xl/data/core\rootview.lua:524: in function 'update'
	D:\pparaf\progs\lite-xl/data/core\rootview.lua:1082: in function 'update'
	D:\pparaf\progs\lite-xl/data/plugins\autocomplete.lua:410: in function 'root_view_update'
	D:\pparaf\progs\lite-xl/data/plugins\contextmenu.lua:26: in function 'root_view_update'
	D:\pparaf\progs\lite-xl/data/plugins\treeview.lua:398: in function 'update'
	D:\pparaf\progs\lite-xl/data/core/init.lua:1529: in function 'step'
	D:\pparaf\progs\lite-xl/data/core/init.lua:1596: in function <D:\pparaf\progs\lite-xl/data/core/init.lua:1592>
	(...tail calls...)
	[string "local core..."]:9: in function <[string "local core..."]:2>
	[C]: in function 'xpcall'
	[string "local core..."]:2: in main chunk
```

The same on Linux (Ubuntu 20.04) and Windows 10

adamharrison:
So we don't crash on a newline being entered. Should close #60 .

### DIFF
--- a/plugins/regexreplacepreview.lua
+++ b/plugins/regexreplacepreview.lua
@@ -124,32 +124,34 @@ local function regex_replace_file(view, pattern, old_lines, raw, start_line, end
     for i = (start_line or 1), (end_line or #doc.lines) do
       local new_text, matches, rmatches
       local old_text = old_lines[i] or doc.lines[i]
-      local old_length = #old_text
-      if replacement then
-        new_text, matches, rmatches = substitute(re, old_text, replacement)
-      end
-      if matches and #matches > 0 then
-        old_lines[i] = old_text
-        replace_line(i, new_text)
-        if line_scroll == nil then
-          line_scroll = i
-          doc:set_selection(i, rmatches[1][1], i, rmatches[1][2])
+      if old_text then
+        local old_length = #old_text
+        if replacement then
+          new_text, matches, rmatches = substitute(re, old_text, replacement)
         end
-      elseif old_lines[i] then
-        replace_line(i, old_lines[i])
-        old_lines[i] = nil
-      end
-      if not replacement then
-        local s,e = regex_match(re, old_text)
-        if s then
-          line_scroll = i
-          doc:set_selection(i, s, i, e)
-          break
+        if matches and #matches > 0 then
+          old_lines[i] = old_text
+          replace_line(i, new_text)
+          if line_scroll == nil then
+            line_scroll = i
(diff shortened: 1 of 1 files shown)


jgmdev:
The version of `regexreplacepreview` should be bumped on the manifest file.

Guldoman:
Just to reiterate what was said on Discord:

The issue is that `raw_remove` correctly removes the newline, so it removes a line from the file. Because the `for` is using the old line count, it goes over.

A simpler solution would be an early exit:
```lua
for i = (start_line or 1), (end_line or #doc.lines) do
  if i > #doc.lines then break end
  -- ...
```
but a better one would be getting the whole file in a buffer and applying the regex there, then replacing the file content with the result, which would also allow multi-line operations.
Getting the whole file in a buffer for "live" operations might be too heavy, so for big files this should probably only be done on the visible part.

Another solution could be ignoring the escapes and just `string.format`ting the input with `%q`.

Last time I tested this plugin I also found out it wasn't updating the undo buffer on apply.

PR #207, opened 2023-01-25
custom_titlebar - A plugin to customize the Lite XL title bar

Minkiu:
A new plugin that allows the person to arbitrary modify the window title bar.

I believe most of the logic is good/correct, but for some reason the title won't change.

If I implement it directly in `data/core/init.lua` https://github.com/lite-xl/lite-xl/blob/master/data/core/init.lua#L1297-L1299 it works, so my question is, how would I go about "overloading" the `core.compose_window_title` ? 

Would that be the best approach?


Delta-official:
just redefine the function like so
```lua
local core = require "core"

function core.compose_window_title(--[[ args here ]])
    -- code here
end
```

jgmdev:
Tested and found that `$savestatus` is not getting replaced to the intended value

Minkiu:
Oh, forgot to remove it from the config text, since the `$filepath` already has that feature, will update now.

PR #208, opened 2023-01-28
Added Haxe language support

AlexRamallo:
Adds syntax highlighting for Haxe `.hx` and `.hxml` files


adamharrison:
Hi, can you add this to the manifest.json?

AlexRamallo:
> Can you add this to the manifest.json? It's used to generate the README, and for plugin managers.

Done

plugins/language_haxe.lua
```diff
@@ -0,0 +1,91 @@
...
+  name = "Haxe",
+  files = "%.hx$",
+  comment = "//",
+  patterns = {
+    { pattern = "%.%.%.",                          type = "operator"   },
+    { pattern = "%<()%u[%w_]*()%>*",               type = {"operator", "keyword2", "operator"}   },
+    { pattern = "%#%s*[%a_]*().*\n",               type = {"keyword", "normal"}   },
+    { pattern = "import%s+()%u[%w]*",              type = {"keyword2", "keyword2"}},
+    { pattern = "import%s+()[%w%.]*%.()%u[%w]*",   type = {"keyword2", "normal", "keyword2"}},
+    { pattern = "//.-\n",                          type = "comment"  },
```
Guldoman, line 35: `//.*` would be better.

plugins/language_haxe.lua
```diff
@@ -0,0 +1,91 @@
+-- mod-version:3
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "Haxe Compiler Arguments",
+  files = "%.hxml$",
+  patterns = {
+    { pattern = "[%-][%-%w_]*",      type="keyword"},
```
Guldoman, line 8: I don't think the square brackets are needed in the first part.

plugins/language_haxe.lua
AlexRamallo: Fixed that, though that was just left over from the C# highlighter which I used as a starting point

plugins/language_haxe.lua
AlexRamallo: Fixed

jgmdev (changes requested):
Thanks a lot for this addition, I remember when I had a crush with the whole Haxe concept of using a single language for multitude of use cases, from scripting to native compilation support :smile: 

Here I add some suggestions that may be applicable as part of the review.

plugins/language_haxe.lua
```diff
@@ -0,0 +1,111 @@
+-- mod-version:3
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "Haxe Compiler Arguments",
```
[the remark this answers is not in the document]
AlexRamallo: Good catch!

plugins/language_haxe.lua
```diff
@@ -0,0 +1,111 @@
...
+    { pattern = "//.*\n",                                   type = "comment"  },
+    { pattern = { "/%*", "%*/" },                           type = "comment"  },
+    { pattern = { '"', '"', '\\' },                         type = "string"   },
+    { pattern = { "'", "'", "\\" },                         type = "string", syntax = ".hx__string_interp"},  
+    { pattern = "-?%.?%d+",                                 type = "number"   },
+    { pattern = "-?0x%x+",                                  type = "number"   },
+    { pattern = "-?%d+%.[%deE]+",                           type = "number"   },
+    { pattern = "-?%d+[%deE]+",                             type = "number"   },
+    { pattern = "[%+%-%.=/%*%^%%<>!~|&]",                   type = "operator" },
+    { pattern = "[%a_][%w_]*%s*%f[(]",                      type = "function" },
```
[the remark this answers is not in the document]
AlexRamallo: This one seems more difficult to deal with. That change fixes the `if (` problem, but breaks highlighting for function calls. I looked at the C++ highlighter to compare, and it seem to have a similar issue, though it opts to break highlighting for `func (` rather than `if (`. I think that's better since it's less common to see the former than the latter.

See the attached screenshots for comparison:

Proposed change:
haxe_1

New version similar to language_cpp:
haxe_2

Unfortunately, it doesn't seem like lite xl gives precedence to keywords when highlighting, so this is probably as good as it'll get for now. (tested on master)

plugins/language_haxe.lua
```diff
@@ -0,0 +1,111 @@
...
+    { pattern = "[%a_][%w_]*",                              type = "symbol"   },
+    { pattern = ":()%u[%a_][%w_]*",                         type = {"normal", "keyword2"}},
+    { pattern = "@:[%a_][%w_]*%f[(]",                       type = "keyword"   },
+    { pattern = "%$type",                                   type = "keyword"   },
+  },
+  symbols = {
```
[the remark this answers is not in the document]
AlexRamallo: I agree. I picked keyword2 because I was using a Sublime Text 3 window to compare while making it, but I just noticed now that the color scheme I typically use there doesn't distinguish between them for some reason, but the others do... so I changed it so `class` `import` and `package` use `keyword`

plugins/language_haxe.lua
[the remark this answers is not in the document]
jgmdev: You should also add:

```diff
files = "%.hxml$",
+ comment = "#",
```

This allows commenting the source code by using the `ctrl + /` binding

plugins/language_haxe.lua
```diff
@@ -54,7 +55,7 @@ syntax.add {
     { pattern = "-?%d+%.[%deE]+",                           type = "number"   },
     { pattern = "-?%d+[%deE]+",                             type = "number"   },
     { pattern = "[%+%-%.=/%*%^%%<>!~|&]",                   type = "operator" },
-    { pattern = "[%a_][%w_]*%s*%f[(]",                      type = "function" },
+    { pattern = "[%a_][%w_]*%f[(]",                         type = "function" },
```
jgmdev, line 58: ahh, my bad I gave you an incorrect pattern before:

wrong one: 
```lua
{ pattern = "[%a_][%w_]()*%s*%f[(]",   type = {"function", "normal"} },
```
good one:
```lua
{ pattern = "[%a_][%w_]*()%s*%f[(]",   type = {"function", "normal"} },
```
Sorry that I made the mistake when adjusting the pattern and left the asterisk as part of the second statement match :sweat: the good one should work:

AlexRamallo: Done!

jgmdev: Thanks a lot!

PR #213, opened 2023-02-13
Fix Rust lifetime highlighting

Delta-official:
before:
Screenshot_20230213_042643

after:
Screenshot_20230213_042635


jgmdev:
The version of `language_rust` should be bumped on the manifest file.

Infinnitor:
I believe the pattern should be changed to `'%a+[%a%d_]*` to allow for lifetime names with numbers and underscores.

Delta-official:
i believe the pattern `[%a_]+[%a%d_]*` would be more correct

adamharrison:
Can you bump the manifest? I'll merge after that happens.

PR #222, opened 2023-03-09
Automatically disabling minimap for non-UTF-8 files

Onlymiind:
Currently if you try to open non-UTF-8 file (e.g. a binary) with minimap enabled, a crash occurs.
These changes automatically disable minimap for non-UTF-8 files.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -949,7 +949,10 @@
       "version": "0.2",
       "path": "plugins/minimap.lua",
       "id": "minimap",
-      "mod_version": "3"
+      "mod_version": "3",
+      "dependencies": {
+        "encoding": {}
+      }
     },
     {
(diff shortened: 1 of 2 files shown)


Guldoman:
This is not the way to go.

The main issue is that the utf-8 library we're using is too strict and just errors out on utf-8 errors (we even have problems in core with it).

A simpler solution would be to just use `string.find` instead of `string.ufind` in https://github.com/lite-xl/lite-xl-plugins/blob/65be429e6f3fc7f8cd265eee3a489357da8a4582/plugins/minimap.lua#L548
but this will also mean that the size of the rects will be a bit off with multi-byte chars.
Maybe a second pass with a more lenient function to determine the number of utf-8 chars should be enough to fix this (if we even care about multi-byte chars accuracy for `minimap`).

PR #229, opened 2023-03-28
Add endwise plugin

LolsonX:
Hi!
I've created a small plugin which allows inserting "end" automatically. It is still in early stages and may not be 100% correct but i think it is good to add it here and eventually update it's version. All improvements suggestions and problems are welcome :)


manifest.json
```diff
@@ -113,6 +113,13 @@
       "id": "datetimestamps",
       "mod_version": "3"
     },
+    {
+      "description": "Add closing end for diffrent languages",
```
Guldoman, line 117: You might want to put the rest of the description you added to the `README` here too, as we generate the `README` from this manifest.

LolsonX: Okay, changed that :)

Guldoman:
How does this compare to `lfautoinsert`?

LolsonX:
> How does this compare to `lfautoinsert`?
If You mean autoinsert plugin then in some languages like ruby there are various rules when end should be applied and when not. For instance
```ruby
if something
  do_something
end
```
You put end here, but when code looks like this
```ruby
do_something if something
```
You don't do this. So the difference between autoinsert and this plugin is that this one takes whole line and look if end should be added, when autoinsert checks only clicked character.

PR #232, opened 2023-04-04
Add snippets plugins

vqns:
Adds code snippets plugins.

I'm not too sure about the [remote manifest](https://github.com/vqns/lite-xl-snippets/blob/main/manifest.json), is it correct?

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -959,6 +959,13 @@
       "id": "lsp_lua"
     },
     {
+      "description": "Adds LSP snippets support to the snippets plugin",
+      "version": "1",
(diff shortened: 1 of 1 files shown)


adamharrison:
Not quite; it's invalid JSON; you're missing a comma after description for the second addon.

Guldoman:
You need to update the git hash here too.

adamharrison:
Installed it, works with lpm. Tried putting in the sample `for` snippet; worked fine. Merging.

Also, one thing, you may want to just adjust your README; the sample with `snippets.add` doesn't have the requisite `require` above it, like the lsp plugin does. Maybe also might want to just mention that people are supposed to add this in their user plugin.

But whatever, minor stuff. Looks good!

PR #233, opened 2023-04-19
Fix MoonScript syntax highlighting

jperon:
@ was causing weird behavior within strings, as illustrated below. I didn’t succeed to get a minimal showcase, so here is the link to [the file that shows the problem](https://github.com/jperon/lyluatex/blob/dev/lyluatex.moon):

without fix:
Capture d’écran du 2023-04-19 09-56-31

with the fix:
Capture d’écran du 2023-04-19 09-57-01

### DIFF
--- a/plugins/language_moon.lua
+++ b/plugins/language_moon.lua
@@ -21,6 +21,7 @@ syntax.add {
     { pattern = "[%a_][%w_]*",            type = "symbol"   },
     { pattern = {"\\", "[%a_][%w_]*"},    type = "function" },
     { pattern = {"%.", "[%a_][%w_]*"},    type = "function" },
+    { pattern = "@[%a_][%w_]*",           type = "keyword2" },
     { pattern = "@",                      type = "keyword2" },
     { pattern = "!",                      type = "keyword2" },
     { pattern = "[%p]",                   type = "keyword"  },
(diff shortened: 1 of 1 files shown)


Guldoman:
The issue is not related to strings, but to the fact that `@` shouldn't start a delimited pattern.
Most likely the correct fix should be something like:
```diff
diff --git a/plugins/language_moon.lua b/plugins/language_moon.lua
index 8042b7d..3ccb7ea 100644
--- a/plugins/language_moon.lua
+++ b/plugins/language_moon.lua
@@ -21,7 +21,8 @@ syntax.add {
     { pattern = "[%a_][%w_]*",            type = "symbol"   },
     { pattern = {"\\", "[%a_][%w_]*"},    type = "function" },
     { pattern = {"%.", "[%a_][%w_]*"},    type = "function" },
-    { pattern = {"@", "[%a_][%w_]*"},     type = "keyword2" },
+    { pattern = "@[%a_][%w_]*",           type = "keyword2" },
+    { pattern = "@",                      type = "keyword2" },
     { pattern = "!",                      type = "keyword2" },
     { pattern = "[%p]",                   type = "keyword"  },
   },
```

jperon:
Change done.

Guldoman:
Thank you!

I'll wait a bit before merging to allow other people to comment.

Guldoman (approved):

PR #234, opened 2023-04-19
language_lilypond

jperon:
This is my first attempt, so feel free to amend!

### DIFF
--- a/language_lilypond.lua
+++ b/language_lilypond.lua
@@ -3,7 +3,7 @@ local syntax = require "core.syntax"
 
 syntax.add {
   name = "LilyPond",
-  files = { "%.ly$" },
+  files = { "%.i?ly$" },
   comment = "%%",
   patterns = {
     { pattern = "#%(()[%a_]%S*",          type = { "operator", "function" } },
--- a/language_lilypond.lua
+++ b/language_lilypond.lua
@@ -1,0 +1,19 @@
+-- mod-version:3
+local syntax = require "core.syntax"
+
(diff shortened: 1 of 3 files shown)


language_lilypond.lua
```diff
@@ -0,0 +1,19 @@
+-- mod-version:3
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "LilyPond",
+  files = { "%.ly$" },
```
Guldoman, line 6: Seems like another possible extension is `.ily`, so you could do:
```suggestion
  files = { "%.i?ly$" },
```

Guldoman (commented):
Thank you for your contribution!

You should also update the `manifest.json` file in the root of this repository.

Seems like `%{` and `%}` are used for block comments, so that should probably be added.

plugins/language_lilypond.lua
```diff
@@ -0,0 +1,21 @@
+-- mod-version:3
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "LilyPond",
+  files = { "%.i?ly$" },
+  comment = "%%",
```
Guldoman, line 7:
```suggestion
  comment = "%%",
  block_comment = { "%%{", "%%}" },
```
This enables the command to add block comments.

jperon:
Should be ready now.

Guldoman (approved):
I'll wait to merge this for a bit, to allow you to add new patterns if anything comes up, and to allow others to comment.

PR #237, opened 2023-04-30
carets-a-lot

ThaCuber:
modified most (if not all) caret related plugins
- motiontrail is now aware of custom_caret, trails' color and shape matches that of cursors'
- motiontrail now works with multiple carets
- custom_caret's block cursor shows character below itself (along with characters around for when moving the cursor between characters)
- motiontrail and smoothcaret now work nicely together
- trail doesn't show when caret is hidden

although, there are some rendering problems with motiontrail that I don't know how to fix, such as it making a trail between two cursors in different docviews or said traill being affected by scrolling

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -195,7 +195,7 @@
     },
     {
       "description": "Replaces selected Lua code with its evaluated result",
-      "version": "0.1",
+      "version": "0.2",
       "path": "plugins/eval.lua",
       "id": "eval",
       "mod_version": "3"
--- a/manifest.json
+++ b/manifest.json
@@ -101,7 +101,7 @@
     },
     {
       "description": "Customize the caret in the editor",
-      "version": "0.2",
+      "version": "0.3",
       "path": "plugins/custom_caret.lua",
       "id": "custom_caret",
       "mod_version": "3"
@@ -1009,7 +1009,7 @@
     },
     {
       "description": "Adds a motion-trail to the caret *([gif](https://user-images.githubusercontent.com/3920290/83256814-085ccb00-a1ab-11ea-9e35-e6633cbed1a9.gif))*",
-      "version": "0.1",
+      "version": "0.2",
       "path": "plugins/motiontrail.lua",
       "id": "motiontrail",
       "mod_version": "3"
@@ -1192,7 +1192,7 @@
     },
     {
       "description": "Smooth caret animation *([gif](https://user-images.githubusercontent.com/20792268/139006049-a0ba6559-88cb-49a7-8077-4822445b4a1f.gif))*",
-      "version": "0.1",
+      "version": "0.2",
       "path": "plugins/smoothcaret.lua",
       "id": "smoothcaret",
       "mod_version": "3"
--- a/plugins/custom_caret.lua
+++ b/plugins/custom_caret.lua
@@ -2,9 +2,9 @@
 
 --[[
   Author: techie-guy
-
   Plugin to customize the caret in the editor
   Thanks to @Guldoman for the initial example on Discord
+  modifications by @thacuber2a03
 
   Features
     Change the Color and Opacity of the caret
@@ -25,6 +25,7 @@ local style = require "core.style"
 local common = require "core.common"
 local config = require "core.config"
 local DocView = require "core.docview"
+
 
 config.plugins.custom_caret = common.merge({
     shape = "line",
@@ -88,16 +89,28 @@ core.add_thread(function()
   end
 end)
 
+local caret_idx = 1
+
+local docview_update = DocView.update
+function DocView:update()
+  docview_update(self)
+  caret_idx = 1
+end
+
 function DocView:draw_caret(x, y)
   local caret_width = style.caret_width
   local caret_height = self:get_line_height()
   local current_caret_shape = conf.shape
   local caret_color = conf.custom_color and conf.caret_color or style.caret
 
+  local font = self:get_font()
+  local line, col = self.doc:get_selection_idx(caret_idx)
+  local charw = math.ceil(font:get_width(self.doc:get_char(line, col)))
+
   if (current_caret_shape == "block") then
-    caret_width = math.ceil(self:get_font():get_width("a"))
+    caret_width = charw
   elseif (current_caret_shape == "underline") then
-    caret_width = math.ceil(self:get_font():get_width("a"))
+    caret_width = charw
     caret_height = style.caret_width*2
     y = y+self:get_line_height()
   else
@@ -106,4 +119,28 @@ function DocView:draw_caret(x, y)
   end
 
   renderer.draw_rect(x, y, caret_width, caret_height, caret_color)
+  if current_caret_shape == "block" then
+    core.push_clip_rect(x, y, caret_width, caret_height)
+
+    local function draw_char(l, c)
+      l = common.clamp(l, 1, #self.doc.lines   )
+      c = common.clamp(c, 1, #self.doc.lines[l])
+      local cx,cy = self:get_line_screen_position(l, c)
+      renderer.draw_text(
+        font, self.doc:get_char(l, c),
+        cx, cy+self:get_line_text_y_offset(),
+        style.background
+      )
+    end
+
+    for yo=-1, 1 do
+      for xo=-1, 1 do
+        draw_char(line+xo, col+yo)
+      end
+    end
+
+    core.pop_clip_rect(x, y, caret_width, caret_height)
+  end
+
+  caret_idx = caret_idx + 1
 end
--- a/plugins/custom_caret.lua
+++ b/plugins/custom_caret.lua
@@ -150,7 +150,7 @@ function DocView:draw_caret(x, y)
       draw_char(line, col)
     end
 
-    core.pop_clip_rect(x, y, caret_width, caret_height)
+    core.pop_clip_rect()
   end
 
   caret_idx = caret_idx + 1
--- a/plugins/motiontrail.lua
+++ b/plugins/motiontrail.lua
@@ -30,43 +30,69 @@ config.plugins.motiontrail = common.merge({
   }
 }, config.plugins.motiontrail)
 
+local cc_installed = pcall(require, 'plugins.custom_caret')
+local cc_conf = config.plugins.custom_caret
 
 local function lerp(a, b, t)
+  if not (a and b and t) then return 0 end
   return a + (b - a) * t
 end
 
+local caret_index = 1
 
 local function get_caret_rect(dv)
-  local line, col = dv.doc:get_selection()
-  local x, y = dv:get_line_screen_position(line, col)
-  return x, y, style.caret_width, dv:get_line_height()
+  local line, col = dv.doc:get_selection_idx(caret_index)
+  local chw = dv:get_font():get_width(dv.doc:get_char(line, col))
+   local w = style.caret_width
+  local h = dv:get_line_height()
+
+  if cc_installed then
+    local cc_shape = cc_conf.shape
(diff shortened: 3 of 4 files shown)


ThaCuber:
that first commit wasn't supposed to be in this one :concerned:

adamharrison:
> that first commit wasn't supposed to be in this one :concerned:

Rebase time!

ThaCuber:
do I need to do it or can you? because if so, idk how

adamharrison:
Lemme see if I can get you a string of the right commands.

adamharrison:
1. `cd /tmp && git clone git@github.com:ThaCuber/lite-xl-plugins.git && cd lite-xl-plugins && git checkout carets-a-lot`
2. `git rebase -i 0768126~`
3. On the first line of the rebase list, change commit 0768126 from `pick` to `drop`.
4. Save.
5. `git push -f`

That should do it.

ThaCuber:
ok, but whi root tmp folder tho

ThaCuber:
ok done

ThaCuber:
wait, force push?

adamharrison:
Yeah; it's because technically it's a rewrite of history on that branch. If you rewrite history, you have to force the push, because by default you shouldn't really ever remove a commit in normal circumstances. It's OK for this though, because it's a transient branch just for a PR.

ThaCuber:
a, ok

can you merge this now e

adamharrison:
I'd have to test it; I'll put it in the queue of things for me to look at soon.

ThaCuber:
just discovered the trail's color doesn't actually change when custom color is deactivated in custom_caret

plugins/motiontrail.lua
```diff
@@ -30,43 +30,69 @@ config.plugins.motiontrail = common.merge({
   }
 }, config.plugins.motiontrail)
 
+local cc_installed = pcall(require, 'plugins.custom_caret')
+local cc_conf = config.plugins.custom_caret
 
 local function lerp(a, b, t)
+  if not (a and b and t) then return 0 end
   return a + (b - a) * t
 end
```
Guldoman, line 39: I think you could replace this with `common.lerp`.

plugins/custom_caret.lua
```diff
@@ -106,4 +119,28 @@ function DocView:draw_caret(x, y)
...
+        cx, cy+self:get_line_text_y_offset(),
+        style.background
+      )
+    end
+
+    for yo=-1, 1 do
+      for xo=-1, 1 do
+        draw_char(line+xo, col+yo)
+      end
+    end
```
Guldoman, line 140: Why do you need to redraw other chars?

Guldoman (commented):
`motiontrail` seems to have problems when the caret blinks and the view is scrolling.

plugins/custom_caret.lua
ThaCuber: it's just for when moving between characters, more of a cosmetic thing than anything else

plugins/motiontrail.lua
ThaCuber: yup

manifest.json
```diff
@@ -195,7 +195,7 @@
     },
     {
       "description": "Replaces selected Lua code with its evaluated result",
-      "version": "0.1",
+      "version": "0.2",
```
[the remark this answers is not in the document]
ThaCuber: why not change the version to show it was modified?

ThaCuber:
> `motiontrail` seems to have problems when the caret blinks and the view is scrolling.

it does, but I have no clue how to solve them
maybe if I moved last_x and last_y to where the caret actually is when it's hidden would work?

manifest.json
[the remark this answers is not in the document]
Guldoman: Because the `eval` plugin isn't modified by this PR.

Guldoman:
The previous version of the plugin was using `DocView:draw` instead of `DocView:draw_caret` so it was running even when the caret was invisible.

plugins/custom_caret.lua
Guldoman: Hopefully it doesn't impact performance much. I'd add a comment about the reason tho.

plugins/custom_caret.lua
```diff
@@ -106,4 +119,28 @@ function DocView:draw_caret(x, y)
...
+
+    core.pop_clip_rect(x, y, caret_width, caret_height)
```
Guldoman, line 142: `core.pop_clip_rect` doesn't take any parameter.

manifest.json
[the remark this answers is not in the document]
ThaCuber: ohhh I didn't notice it was the eval plugin lol, yes it was a mistake

plugins/custom_caret.lua
ThaCuber: I don't remember putting those there...

ThaCuber:
> The previous version of the plugin was using `DocView:draw` instead of `DocView:draw_caret` so it was running even when the caret was invisible.

problem with that is, the trail shows even with the caret hidden if using `DocView:draw`, and using `DocView:draw` instead of `DocView:draw_caret` doesn't give me the coordinates of the caret at the moment, which makes it not play well with smoothcaret

plugins/motiontrail.lua
```diff
@@ -30,43 +30,69 @@ config.plugins.motiontrail = common.merge({
...
 
-  if last_view == self and (x ~= last_x or y ~= last_y) then
+  if lsw == self and (x ~= lsx or y ~= lsy) then
     local lx = x
     for i = 0, 1, 1 / config.plugins.motiontrail.steps do
-      local ix = lerp(x, last_x, i)
-      local iy = lerp(y, last_y, i)
+      local ix = lerp(x, lsx, i)
+      local iy = lerp(y, lsy, i)
+      if cc_conf.shape == "underline" then iy = iy + self:get_line_height() end
```
Guldoman, line 85: You should check if the plugin is actually installed, otherwise if the shape is `underline` but `custom_caret` isn't active, the trail isn't correct.

Guldoman:
You could check if multiple `DocView:draw` happened before the `draw_caret`, and in that case don't draw the trail.

ThaCuber:
> You could check if multiple `DocView:draw` happened before the `draw_caret`, and in that case don't draw the trail.

how?

plugins/custom_caret.lua
ThaCuber: even better, I made it a setting
Captura desde 2023-05-04 22-49-47

Guldoman:
Something like
```lua
function DocView:draw()
  self.draws = self.draws + 1
end

function DocView:draw_caret(...)
  -- ...
  if self.draws <= 1 then
    -- draw trail here
  end
  -- ...
  self.draws = 0
end
```

ThaCuber:
that didn't exactly work

~~also don't make them object variables, it feels wrong~~

plugins/motiontrail.lua
```diff
@@ -30,43 +30,71 @@ config.plugins.motiontrail = common.merge({
...
-      local iy = lerp(y, last_y, i)
+      local ix = common.lerp(x, lsx, i)
+      local iy = common.lerp(y, lsy, i)
+      if cc_installed and cc_conf.shape == "underline" then
+        iy = iy + self:get_line_height()
+      end
       local iw = math.max(w, math.ceil(math.abs(ix - lx)))
-      renderer.draw_rect(ix, iy, iw, h, style.caret)
+      local color = style.caret
+      if cc_conf.custom_color then
```
Guldoman, line 86: Check if the plugin is installed here too.

Guldoman:
> ~also don't make them object variables, it feels wrong~

You can use a *weak* table for that in the plugin itself.

Guldoman:
> that didn't exactly work

You still need to update the `last_{x,y}` coordinates, even if you didn't draw.

ThaCuber:
> You still need to update the `last_{x,y}` coordinates, even if you didn't draw.

image

Guldoman:
```diff
diff --git a/plugins/motiontrail.lua b/plugins/motiontrail.lua
index e395296..416134d 100644
--- a/plugins/motiontrail.lua
+++ b/plugins/motiontrail.lua
@@ -61,6 +61,12 @@ function DocView:update()
   dv_update(self)
 end
 
+local dv_draw = DocView.draw
+function DocView:draw(...)
+  self.draws = self.draws and self.draws + 1 or 1
+  return dv_draw(self, ...)
+end
+
 local dv_draw_caret = DocView.draw_caret
 function DocView:draw_caret(x, y)
   if not config.plugins.motiontrail.enabled or self ~= core.active_view then
@@ -68,6 +74,9 @@ function DocView:draw_caret(x, y)
     return
   end
 
+  if self.draws > 1 then
+    goto skip_draw
+  end
   local lsx, lsy = last_x[caret_idx], last_y[caret_idx]
   local w, h = get_caret_size(self, caret_idx)
 
@@ -92,9 +101,12 @@ function DocView:draw_caret(x, y)
     core.redraw = true
   end
 
+  ::skip_draw::
+
   last_x[caret_idx], last_y[caret_idx], last_view[caret_idx] = x, y, self
 
   caret_idx = caret_idx + 1
+  self.draws = 0
 
   dv_draw_caret(self, x, y)
 end
```

ThaCuber:
wrapped the draw code inside a do end block because the plugin threw an error with reason "jumping into local lsx's scope" but hey, it worked :)

Guldoman:
I edited the comment to fix that.
Tbh I just used the goto to make the diff easier to read.

ThaCuber:
now, about the carets appearing from nowhere after you make some (carets), destroy them and then try to make new ones...

ThaCuber:
I'll just keep track of the currently used carets and reset the last_x / last_y of the unused ones to the new caret position

ThaCuber:
this worked, but I'm scared I might have just swept the problem under the rug...
```lua
local caret_idx, caret_amt = 1, 0

local dv_update = DocView.update
function DocView:update()
  self.last_x = self.last_x or {}
  self.last_y = self.last_y or {}
  self.last_view = self.last_view or {}
  caret_idx = caret_idx or 1

  -- continue from whatever caret_idx left
  caret_amt = caret_amt and math.max(caret_amt, caret_idx) or 0
  for i=1, caret_amt - caret_idx do
    self.last_x[caret_idx + i], self.last_y[caret_idx + i], self.last_view[caret_idx + i] = nil, nil, nil
  end
  caret_idx = 1
  dv_update(self)
end
```

ThaCuber:
plus, it still has the original problem it had when it was first uploaded, the cursor trail appears when scrolling

ThaCuber:
> this worked, but I'm scared I might have just swept the problem under the rug...
>
> (code block)

I'll commit this while I think about the rest

Guldoman:
> plus, it still has the original problem it had when it was first uploaded, the cursor trail appears when scrolling

You could check if the actual line-column position changed. Either by using the coordinates to get the line-col, or by checking the selections.

ThaCuber:
alr, this should be everything now

plugins/custom_caret.lua
adamharrison: Could you take'em out?

adamharrison:
Looks good to me. Tested it; changed cursors, checked on the bugs mentioned, everything seems resolved to me (even if the clearing out of caret info is a bit sketch). As long as we remove those parameters on the pop, I'd be good to merge this.

ThaCuber:
what the hell wdym they're still there 💀

ThaCuber:
also, well, you see, the block caret is using the character inmediately at the cursor position, and only that, which means it doesn't really take into account unicode...

PR #240, opened 2023-05-03
Added in `www` to manifest.

adamharrison:
Adds in a library for HTTP(s) requests.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1376,7 +1376,7 @@
       "description": "A simple library that provides a web client to fetch, and submit data via HTTP requests.",
       "version": "0.1",
       "type": "library",
-      "remote": "aecc866c33b0a954fbd41a5de2842bb97f4ac45e",
+      "remote": "https://github.com/adamharrison/lite-xl-www.git:aecc866c33b0a954fbd41a5de2842bb97f4ac45e",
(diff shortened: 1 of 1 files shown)


manifest.json
```diff
@@ -1370,6 +1370,14 @@
       "path": "plugins/wordcount.lua",
       "id": "wordcount",
       "mod_version": "3"
+    },
+    {
+      "id": "www",
+      "description": "A simple library that provides a web client to fetch, and submit data via HTTP requests.",
+      "version": "0.1",
+      "type": "library",
+      "remote": "https://github.com/adamharrison/lite-xl-www.git:aecc866c33b0a954fbd41a5de2842bb97f4ac45e",
```
Guldoman, line 1379: Could the `.git` part of the URL cause issues?

adamharrison: Don't see how; that is part of the official URL as provided by github.

Guldoman: It's just that some plugins are using `.git`, others aren't. I guess it's fine.

adamharrison: Odd. Github probably handles it the same either way, if it starts getting requests from a git client. But technically, the `.git` is official.

PR #247, opened 2023-05-05
[Plugin] Overwrite

ThaCuber:
it's strange that I haven't yet seen a plugin like this, so I made one
it lets you toggle between inserting characters and overwriting them
I'm making many PRs lately, hope it's not an issue...


adamharrison:
Not an issue at all. Just tested it. Works great on single cursor, and multi-cursor. Let me just review the code, and then we'll probably be good to go.

Guldoman:
This should probably be a core feature.

adamharrison:
Yeah; honestly, looks great. Merging.

adamharrison:
As per deicison on github, this should be submitted as a core change. Closing.

PR #248, opened 2023-05-15
rename treeview-menu-extender to treeview-extender

juliardi:


Guldoman:
Would you mind updating the `manifest.json`?

juliardi:
ok, coming soon

juliardi:
@Guldoman `manifest.json` added

juliardi:
@Guldoman `manifest.json` updated

manifest.json
```diff
@@ -275,7 +275,7 @@
     {
       "description": "Shows \"git blame\" information of a line *([screenshot](https://raw.githubusercontent.com/juliardi/lite-xl-gitblame/main/screenshot_1.png))*",
       "version": "0.1",
```
Guldoman, line 277: Should probably bump this too.

Guldoman (commented):
I think you should add a `"replaces": [ "treeview-menu-extender" ]` field, so that people will automatically use the new version.
@adamharrison is that the correct syntax/usage?

manifest.json
adamharrison: > I think you should add a `"replaces": [ "treeview-menu-extender" ]` field, so that people will automatically use the new version. @adamharrison is that the correct syntax/usage?

This is the correct syntax, yes.

juliardi: @Guldoman But there is no newer version for `gitblame` though.

Guldoman: You changed the commit it was pointing to, so it should be bumped.

juliardi: alright, it's done. can you check it please?

PR #253, opened 2023-05-22
Add Nonicons font as a library and make the `nonicons` plugin use it

Guldoman:
This should allow other plugins to use the Nonicons font more easily.
The mapping should make interacting with the font a bit simpler.

A drawback of this change is that people that installed the font in `USERDIR/fonts/nonicons.ttf` need to move it in `USERDIR/libraries/font_nonicons/nonicons.ttf`.

People that use `lpm` should get the font automatically.


Guldoman:
Rebased and updated.

Any comments on the drawback mentioned in the initial message?
Comments on how the mapping is handled?

Will merge in a few days if there are no objections.

adamharrison:
Tested, works for me (minus some lpm fuckery).

Guldoman:
Moved the font library from a directory to a file.
It now also returns the path of the font.

PR #254, opened 2023-05-22
Add NerdFont Symbols font as a library and make the `nerdicons` plugin use it

Guldoman:
Same drawbacks and advantages as #253.

This also replaces some of the icons that were removed from NerdFont.


Gaspartcho:
Hello, is this PR still being considered to be merged? I wanted to open one in order to fix the icons of `nerdicons`, but as I saw in #287, it seems that this PR needs to be merged before anything.

Guldoman:
Rebased and updated.
Same caveats as #253.

Let me know if the `mapping` approach makes sense.
It should help when updating to new Nerd Font versions, as well as improves readability and reusability of the font.

thacuber2a03:
why can't you move the font file from `libraries/font_*icons/*icons.ttf` to `libraries/*icons.ttf`?

Guldoman:
Because it's part of the `font_symbols_nerdfont_mono_regular` library.
I could make that library just return the path for the font tho.

thacuber2a03:
also, about the mapping approach, finding the name and using it in the library is really convenient, and you can retype it if you accidentally delete the icon, and moving files and folders around isn't too slow either (unless maybe you have an SSD or you don't know where your `USERDIR` is for whatever reason), so really I'm fine with this choice

Guldoman:
Moved the font library from a directory to a file (the installed ttf will still be inside the subdirectory).
It now also returns the path of the font.

PR #256, opened 2023-05-25
Fix indentation in `lfautoinsert` and add support for multiple carets

Guldoman:
This also avoids using `command.perform` as it can be problematic (for example `doc:newline` behaves in different ways depending on `config.keep_newline_whitespace`).

Needs testing.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -931,7 +931,7 @@
     },
     {
       "description": "Automatically inserts indentation and closing bracket/text after newline",
-      "version": "0.1",
+      "version": "0.2",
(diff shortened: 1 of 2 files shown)


adamharrison:
Just tested this. Looks good. Tested with multiple cursors, multiple different types of autoinsert; tested that it doesn't insert extra tags if there is one.

I did however manage to get into a weird state where it wasn't auto inserting anything. Will do more testing to see if I can repro.

adamharrison:
Not a weird state; intended behaviour with extraneous indents below the opening tag. Merging.

PR #258, opened 2023-06-17
fixes and improves plugin dragdropselected

SwissalpS:
- Tighter and neater code
- Removed weird selection during drag
- Left-ctrl+drag works if gnome's "ctrl to show mouse" is turned off
- Added sticky mode to help against finger fatigue
- Copied selections can now be placed adjacent to the original selection
- Press Esc to abort a begun action
- Added toggles for settings GUI to enable/disable plugin and sticky mode during runtime

Edit:
- Selects copied/moved portion after action
- Updates manifest.json

Edit2:
- Multiple selections are handled

Edit3:
- User can optionally have a ghost of selected text follow mouse when dragging (requires widget library)
- User can limit the length of the ghost.


plugins/dragdropselected.lua
```diff
@@ -29,149 +60,171 @@ local function isInSelection(iLine, iCol, iSelLine1, iSelCol1, iSelLine2, iSelCo
...
-  if sCursor then self.cursor = sCursor end
-
+    self.cursor = 'hand'
+    -- make sure selection is marked
+    self.doc:set_selection(table.unpack(self.dnd_lSelection))
+  end
+  -- calculate line and column for current mouse position
+  local iLine, iCol = self:resolve_screen_position(x, y)
+  -- show insert location (unfortunately it doesn't always show, even when
+  -- calling draw_caret)
```
Guldoman, line 86: You could set `config.disable_blink` (and restore it later).

Guldoman (commented):
Could you bump the plugin version in the manifest?

plugins/dragdropselected.lua
SwissalpS: Thanks, I'll try that.

SwissalpS: works like a charm

manifest.json
```diff
@@ -142,8 +142,8 @@
       "id": "discord-presence"
     },
     {
-      "description": "Provides basic drag and drop of selected text (in same document)",
-      "version": "0.2",
+      "description": "Provides drag and drop of selected text (in same document)",
+      "version": "20230616.094245",
```
Guldoman, line 146: @adamharrison could you confirm that this is a valid version for lpm?

SwissalpS: yes it does:
https://github.com/lite-xl/lite-xl-plugin-manager/blob/master/SPEC.md#:~:text=version%3A%20The%20addon%27s%20semantic%20version.%20A%20string%20that%20can%20contains%20%5B0%2D9%5C.%5D.

Guldoman: I mean with an `lpm` - the software - perspective. Iirc it's slightly limited in that regards.

SwissalpS: spec should state that if there are limits other than characters used.

I have asked: https://github.com/lite-xl/lite-xl-plugin-manager/issues/22

adamharrison: It will work, but isn't convention of major.minor.revision.

SwissalpS: neither this nor most plugins fit into a major.minor.patch scenario :D

adamharrison: minor and patch are optional; they're assumed to be 0 if omitted.

plugins/dragdropselected.lua
```diff
@@ -1,177 +1,263 @@
...
-                                             iSelLine2, iSelCol2)
-  if self.bClickedIntoSelection then
-    self.drag_start_loc = { x, y }
-    -- stash selection for inserting later
-    self.sDraggedText = self.doc:get_text(self.doc:get_selection())
-    self.dragged_selection = { iSelLine1, iSelCol1, iSelLine2, iSelCol2 }
-  else
-    self.bClickedIntoSelection = nil
-    self.dragged_selection = nil
+  local iLine1, iCol1, iLine2, iCol2 = self.doc:get_selection(true)
```
Guldoman, line 137: Save also the swapped status
```suggestion
  local iLine1, iCol1, iLine2, iCol2, swapped = self.doc:get_selection(true)
```
and pass it to `doc:set_selection`, so that the selection "order" is preserved.

Guldoman (commented):
I wonder if you could manage multiple selections, so that it uses the one that's under the mouse and not just the "first" one.

plugins/dragdropselected.lua
SwissalpS: no, it didn't help https://github.com/SwissalpS/liteXLplugins/tree/fixDragDropSelectedSwapped

I have started work on multi-selection scenarios though ;)

Guldoman: What I meant is the location of the caret in relations to the selection:
Schermata del 2023-06-19 16-37-50
vs
Schermata del 2023-06-19 16-37-56
both when starting the drag, and when applying it.

And your changes seem to fix this.

SwissalpS:
I noticed two things while working on this:
1. sometimes when a lot of scrolling is involved, the dragged selection isn't dropped where user would expect. Sometimes many lines off. Happens both when dragging or using sticky mode.
2. ``Doc:get_selection_text()`` returns each part on its own line. Mostly that is fine. In below scenario, user might be a bit surprised:
Captura de pantalla de 2023-06-19 20-44-51
to get:
Captura de pantalla de 2023-06-19 20-46-27
instead of:
Captura de pantalla de 2023-06-19 20-46-57

IDK what would break if ``Doc:get_selection_text()`` was changed to return items on same lines delimited with spaces instead of NL.

Guldoman:
> 1. sometimes when a lot of scrolling is involved, the dragged selection isn't dropped where user would expect. Sometimes many lines off. Happens both when dragging or using sticky mode.

What do you mean? I can drag a word from the end of an 800 lines doc to the first page without issues.

> 2. `Doc:get_selection_text()` returns each part on its own line. Mostly that is fine. In below scenario, user might be a bit surprised:

You could either implement your own `get_selection_text`, or you could replace newlines with spaces in the returned string.

SwissalpS:
> What do you mean?

I don't know how to explain it more clearly. Sometimes dragged text does not insert where mouse was released.

> You could either implement your own get_selection_text

sure, I 'could', question is, would it break anything if that was default behaviour.

> you could replace newlines with spaces

that won't work, users who have used this feature in other text editors expect items from different lines to appear on different lines and items from same lines to be separated by spaces. (At least that's what I remember some of my previous editors to do)

Guldoman:
> > You could either implement your own get_selection_text
> 
> sure, I 'could', question is, would it break anything if that was default behaviour.
> 
> > you could replace newlines with spaces
> 
> that won't work, users who have used this feature in other text editors expect items from different lines to appear on different lines and items from same lines to be separated by spaces. (At least that's what I remember some of my previous editors to do)

I mean, just create your custom function that has the enhanced functionality you want `get_selection_text` to have.

Guldoman:
Would you merge the `swapped` commit too?

SwissalpS:
I don't intend to. Didn't have any effect and now the code scans all selections anyway.

Am I misunderstanding something?

Guldoman:
[Video del 2023-06-20 02-35-04.webm](https://github.com/lite-xl/lite-xl-plugins/assets/2798487/6dffbc54-d21e-4578-aca0-2ef8b196154c)

Also seems like this now *adds* a caret when you release on the selection itself.

SwissalpS:
Unfortunately I can't get any screen-grabs to work on this machine.
- disable ``dragdropselection`` plugin
- select with mouse/keyboard from right to left -> at end of selection (on left) there is a caret
- select from left to right -> at end of selection (on right) there is a caret

To me this seems reasonable and wanted behaviour. It lets user know which way they were selecting even if they were interrupted for a long period.

This plugin does not change any of that.  
When dropping dragged text, caret is at the position where mouse was released -> top left of new selection. This also is wanted behaviour.  
Picture this scenario:
- User selects 42+ times his screen height worth of text
- Drags it to the insertion point and releases

If the caret were to be placed at end of selection, this would trigger a massive scroll event or an abrubt jump to end of pasted text. This may be wanted when using keyboard ctrl-c & ctrl-v to paste text. Here user knows exactly where insertion point is. 
With mouse however, user wants a confirmation that they didn't slip while releasing.
Having document stay put where user let go of mouse is better UX IMO.
User can then with one keystroke arrive at end of selection (right arrow), if that is what they want.

I'm not sure I understood your issue with this behaviour, so I'm trying to elaborate and figure it out. To me it feels natural and correct.   
If I'm missing your point, be so kind and try to describe your issue in more detail.

> Also seems like this now _adds_ a caret when you release on the selection itself.

I have debugged that and confirmed that there is no extra caret added.
https://github.com/SwissalpS/liteXLplugins/tree/fixDragDropSelectedDebugCaret

plugins/dragdropselected.lua
SwissalpS: not using ``Doc:get_selection()`` anymore. Wasn't accurate when user had multiple selections.

Guldoman:
Look at the video. I select left to right, but the instant I start the grab, the selection goes right to left.

> I have debugged that and confirmed that there is no extra caret added.

I can't test again at the moment. Will test again later and post a video if I can still reproduce the issue.

SwissalpS:
> Look at the video. I select left to right, but the instant I start the grab, the selection goes right to left.

Yes, I don't see anything wrong with that.  
The user is made aware of what to expect selection to look like when they release.

Guldoman:
Every other text editor I tested puts the caret at the end, but most importantly they don't change the orientation of the selection when the drag starts.

You're thinking about the user wanting to put the selection *after* something, but the whole discourse is valid also when the user wants to put the selection *before* something. That's why I think leaving the caret where the user left it is the better UX.

About the added selection bug:

[Video del 2023-06-21 06-17-25.webm](https://github.com/lite-xl/lite-xl-plugins/assets/2798487/ccae7af5-0f9c-40cb-88ee-e12a4d993c1c)

SwissalpS:
> Every other text editor I tested puts the caret at the end

Some do, yes and it is annoying but not as annoying as those that remove caret entirely.

> You're thinking about the user wanting to put the selection _after_ something,

nope, I'm thinking of inserting above, below or between selections. In particular how page scroll update can be confusing. Leaving carret at beginning of inserted portion is calmest.

> About the added selection bug:

Thanks for clarifying, I'll look into that, maybe I didn't test non-sticky mode enough.

SwissalpS:
OK, stray caret is gone and carets of selections are preserved where user put them, but I'm not going to scroll to end of newly inserted selection. If user wants to jump there, he can do so with one keypress.

Guldoman:
Last few things: should cancelling the drag with escape keep the original selection?
It might be a good idea to have the escape action a command, mostly to allow changing the keybind. I know keymaps are on keypress and not on release, but it should be fine for this use-case.
Also we should probably disallow writing while dragging, or we should cancel the drag on text input.

SwissalpS:
thanks
- the escape idea is good will look into it
- disabling text input during drag, yes. It'll have to allow keystrokes though or escape won't work
- a friend told me to add setting as to where user wants the carret to be after insert. I'm not sure it's worth the extra code. I would add three options: a) at top-left (current) b) at bottom-right (like libre office) c) use the position of the dragged portion. What do you think about that?

Guldoman:
> * disabling text input during drag, yes. It'll have to allow keystrokes though or escape won't work

Overriding `DocView:on_text_input` might be enough.

> * a friend told me to add setting as to where user wants the carret to be after insert. I'm not sure it's worth the extra code.

I'm fine with it. In the end it's your choice tho.

SwissalpS:
I've implemented the suggestions.  
There is a side-effect though to having the abort procedure be a command: now user can't escape out of autocomplete suggestions :/
[to test use this branch](https://github.com/SwissalpS/liteXLplugins/tree/fixDragDropSelectedMakeEscapeAcommand)

There is one other quirk I want to try and fix in this PR
When clicking on the right side of a selection (multiline shows it best), way out in non-text area, the click is registered as in the selection.  
This is only a problem with sticky mode enabled.

Guldoman:
> There is a side-effect though to having the abort procedure be a command: now user can't escape out of autocomplete suggestions :/ [to test use this branch](https://github.com/SwissalpS/liteXLplugins/tree/fixDragDropSelectedMakeEscapeAcommand)

Instead of using the predicate `"'core.docview'"`, use a function that checks `core.active_view.dnd_sText` or something like that.

SwissalpS:
yes, thanks, I just figured that out too :)

SwissalpS:
offtopic: while tracking down the error, I noticed that disabling the ``contextmenu`` plugin does not seem to have any effect at all (after re-launching lite-xl)

SwissalpS:
just noticed something I want to tweak before merge

SwissalpS:
Is there already a mechanism implemented in lite-xl to allow system-wide drag events?
All I have seen is file-dropping into lite-xl.

In other projects it was easiest to implement inter-view dragging by using system events. Saves time maintaining code for the two distinct event types.

If there is no system-wide drag event built in or planed to be built in, I might find the time to refactor the code to work between views of the same lite-xl instance.

Guldoman:
We plan to have an internal dnd system (I have a prototype already, next week I'll finally have time to dedicate to it), and eventually make it support system-wide events.

For the system-wide part, SDL only supports receiving file dropped events iirc, so I'll have to open a feature request/PR for SDL to implement the rest.
That'll be exclusive to SDL3, so to actually make use of it we'll have to first move to SDL3 (which is also planned).

SwissalpS:
After 1.5 years I see that this repo still has a malfunctioning version of my plugin.
I've resolved the conflicts and this is again ready to be merged.

Please merge this PR or remove the plugin from this repo. I'd rather not have it in here if it's not in working condition.

adamharrison:
Sorry about the delay. We're a bit understaffed, but you're right, 1.5 years is probably too long. (I also have PRs that old which are languishing a bit).

I just tested, and it seems good. There are some points that weren't addressed, but it's better than what's there now, and, as you noted, it is your plugin. I'm merging this now.

SwissalpS:
Many thanks.

PR #262, opened 2023-06-21
make copyfilelocation work on the treeview

ThaCuber:
you can now right click a file in the treeview and the option to copy it's path will be there
image

### DIFF
--- a/plugins/copyfilelocation.lua
+++ b/plugins/copyfilelocation.lua
@@ -1,15 +1,35 @@
 -- mod-version:3
 local core = require "core"
 local command = require "core.command"
+local treeview = require 'plugins.treeview'
+local menu = treeview.contextmenu
 
-command.add("core.docview", {
-  ["copy-file-location:copy-file-location"] = function(dv)
-    local doc = dv.doc
+command.add("core.docview!", {
+  ["doc:copy-file-location"] = function(view)
+    local doc = view.doc
     if not doc.abs_filename then
       core.error "Cannot copy location of unsaved doc"
       return
     end
-    core.log("Copying to clipboard \"%s\"", doc.abs_filename)
+    core.log("Copying \"%s\" to clipboard", doc.abs_filename)
     system.set_clipboard(doc.abs_filename)
   end
+})
+
+command.add(function()
+  return treeview.hovered_item ~= nil
+end, {
+  ["treeview:copy-file-location"] = function()
+    local item = core.active_view.hovered_item
+    if not (item and item.abs_filename) then
+      core.error "Cannot copy location of item"
+      return
+    end
+    core.log("Copying \"%s\" to clipboard", item.abs_filename)
+    system.set_clipboard(item.abs_filename)
+  end
+})
+
+menu:register(nil, {
+  { text = "Copy file location", command = "treeview:copy-file-location" }
 })


plugins/copyfilelocation.lua
```diff
@@ -1,15 +1,35 @@
...
 local core = require "core"
 local command = require "core.command"
+local treeview = require 'plugins.treeview'
+local menu = treeview.contextmenu
 
-command.add("core.docview", {
-  ["copy-file-location:copy-file-location"] = function(dv)
-    local doc = dv.doc
+command.add("core.docview!", {
+  ["doc:copy-file-location"] = function(view)
```
Guldoman, line 8: Ideally the command name should keep the name of the plugin.

plugins/copyfilelocation.lua
```diff
@@ -1,15 +1,35 @@
...
     system.set_clipboard(doc.abs_filename)
   end
 })
+
+command.add(function()
+  return treeview.hovered_item ~= nil
+end, {
+  ["treeview:copy-file-location"] = function()
+    local item = core.active_view.hovered_item
+    if not (item and item.abs_filename) then
```
Guldoman, line 24:
```suggestion
command.add(function()
  return treeview.hovered_item ~= nil, treeview.hovered_item
end, {
  ["treeview:copy-file-location"] = function(item)
    if not (item and item.abs_filename) then
```

plugins/copyfilelocation.lua
```diff
@@ -1,15 +1,35 @@
...
+      core.error "Cannot copy location of item"
+      return
+    end
+    core.log("Copying \"%s\" to clipboard", item.abs_filename)
+    system.set_clipboard(item.abs_filename)
+  end
+})
+
+menu:register(nil, {
```
Guldoman, line 33: Might be a good idea to check in the predicate if the user is hovering an item, to avoid showing the entry when right-clicking on empty space.

plugins/copyfilelocation.lua
```diff
@@ -1,15 +1,35 @@
...
-    core.log("Copying to clipboard \"%s\"", doc.abs_filename)
+    core.log("Copying \"%s\" to clipboard", doc.abs_filename)
     system.set_clipboard(doc.abs_filename)
   end
 })
+
+command.add(function()
+  return treeview.hovered_item ~= nil
+end, {
+  ["treeview:copy-file-location"] = function()
```
Guldoman, line 22: Having the plugin name here too, might be a good idea.

plugins/copyfilelocation.lua
ThaCuber: but there's an equivalent for the treeview too, do I just merge them?

plugins/copyfilelocation.lua
ThaCuber: damn, that just works??

plugins/copyfilelocation.lua
ThaCuber: but doesn't that overwrite the previous command of the same name?

Guldoman: You could call it `copy-file-location:copy-treeview-item-location` or something like that.

plugins/copyfilelocation.lua
Guldoman: Yep, neat system introduced to avoid having to possibly repeat the same work done by the predicate.

We really need docs...

plugins/copyfilelocation.lua
Guldoman: Eh, they're doing different things, so it makes sense for them to be separate.

plugins/copyfilelocation.lua
SwissalpS: with current proposal, context-menu-entry only shows when a file is hovered. -> resolved

ThaCuber:
wait, hold up, let me do one last thing

ThaCuber:
k done

SwissalpS:
I think it would be nice if the commands reflected which plugin added them:
``copyfilelocation:copy-doc-view-file-location`` and ``copyfilelocation:copy-tree-view-file-location``  
or ``copy-file-location:....`` or something of that sort.

ThaCuber:
for the last suggestion: YES

ThaCuber:
how bout
image
image

ThaCuber:
mmmmm
I'm starting to think I could instead make a version of this that lets you also copy the path of a file and rename it to "copylocation.lua"

SwissalpS:
hmm, looking at it there, it would be better English:
"Copy File Location Of Doc(ument)" and "Copy File Location From Tree View" though the later won't show up as it can only be used with mouse highlighting something.

SwissalpS:
> a file and rename it to "copylocation.lua"

I see your point, but don't think that's a good move. But not too bad either.

SwissalpS:
> "copylocation.lua"

why not "copypath"? It copies paths of directories and files from treeview and path of open file.

ThaCuber:
yeah that makes more sense

SwissalpS:
about renaming:  
some of my initial concerns don't apply to this plugin as there are no settings bound to it.

Renaming could lead to some confusion with users and maintainers. Some will have the old one still installed and maybe see the new one, try it out and eventually figure out, that this is the same one with improvements.

Otherwise, yes it does seem to be worth considering and discussing.


Something I'm not clear about yet, is how the handling of disabled plugins is to be done.
Is it a bug in core or are the plugins supposed to check if another plugin they are loading is activated.  
It seems that user can deactivate ``plugins.contextmenu`` for example, but because another plugin requires it, it gets loaded anyway.

In either case, this plugin should not add menu-items if user has contextmenu disabled.

ThaCuber:
well what about not renaming, but releasing under a different name

SwissalpS:
I doubt the repo maintainers will keep both, so essentially the same outcome as renaming

SwissalpS:
I've suggested how to check user settings in case user does not have plugins activated that this plugin manipulates.
https://github.com/ThaCuber/lite-xl-plugins/pull/1

PR #263, opened 2023-06-23
ghmarkdown add options

SwissalpS:
- give user the power to easily change which application to use as they
may not want to use their default browser or do not want to fill history
with temporary files.

e.g. change browser to specifically chromium-freeworld and open in a new
incognito window:
``chromium-freeworld --incognito --new-window %q``

- allow user to set the URL to POST the request to. They may have their
own local processor.

- I've opend as draft-PR as I'm not sure my wording in the settings are well formulated


adamharrison:
This seems reasonable. Tested this, works. I've also bumped the version for you in the manifest. Make sure that any time you change the plugin, you also bump the version.

PR #266, opened 2023-06-27
Update settings.lua max log items default is 800 not 80

SwissalpS:
see: https://github.com/lite-xl/lite-xl/blob/cf028c510a908a1e208588ec90b57fef422f88bf/data/core/config.lua#L6

### DIFF
--- a/plugins/settings.lua
+++ b/plugins/settings.lua
@@ -618,7 +618,7 @@ settings.add("Development",
       description = "The maximum amount of entries to keep on the log UI.",
       path = "max_log_items",
       type = settings.type.NUMBER,
-      default = 80,
+      default = 800,
       min = 50,
       max = 2000
     },


Guldoman:
Would you mind bumping the version in the manifest too?

SwissalpS:
I also suggest setting minimum to something around 150.  
When setting it to 10 it clamped at 132 entries on my test-setup. There are already that many entries before first check.

Options:
- rework core to not bypass check
- rework core to purge overflow when checking
- just ignore and set minimum in this plugin to something like 150-200

SwissalpS:
> Would you mind bumping the version in the manifest too?

sure, thanks, I totally forgot.

Guldoman:
> When setting it to 10 it clamped at 132 entries on my test-setup. There are already that many entries before first check.

This is likely a `settings` issue, as it loads the config value after a few logs are already added.
But then one could argue that this is a core issues, as it should respect the new value anyways.

> I also suggest setting minimum to something around 150.

I'd be fine with it. Smaller values don't make much sense.

PR #267, opened 2023-06-29
respect user setting to disable contextmenu

SwissalpS:


SwissalpS:
It's been about 1.5 years. Why has this neither been merged nor rejected?
Please let me know what I need to change to have this merged.

PR #268, opened 2023-07-18
remember that easing previewer plugin? yeah, I finally uploaded it

ThaCuber:
[in case you don't get what I mean](https://www.youtube.com/watch?v=y6_Prq8hRGE)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -151,7 +151,7 @@
     {
       "description": "Adds a popup that displays the curve of Penner-styled easing functions.",
       "version": "1.0",
-      "remote": "https://github.com/ThaCuber/lite-xl-easingpreview",
+      "remote": "https://github.com/ThaCuber/lite-xl-easingpreview:ea349d12a9570264408dc1567af660b8d4568d3d",
       "id": "easingpreview",
       "mod_version": "3"
     },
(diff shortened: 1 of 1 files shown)


manifest.json
```diff
@@ -148,6 +148,13 @@
       "id": "dragdropselected",
       "mod_version": "3"
     },
+    {
+      "description": "Adds a popup that displays the curve of Penner-styled easing functions.",
+      "version": "1.0",
+      "remote": "https://github.com/ThaCuber/lite-xl-easingpreview",
```
Guldoman, line 154: You must specify a commit hash like this:
```suggestion
      "remote": "https://github.com/ThaCuber/lite-xl-easingpreview:ea349d12a9570264408dc1567af660b8d4568d3d",
```

Might also be worth it to add a `name` field.

ThaCuber: whi though (the commit hash)

Guldoman: To pin the version to the commit (and as a way to avoid malicious changes).

ThaCuber: also, can I write "Easing Previewer" for the `name` field?

Guldoman: Yep.

PR #272, opened 2023-07-30
Added code-plus plugin details to manifest.json

chqs-git:


manifest.json
```diff
@@ -1417,6 +1417,13 @@
       "type": "library",
       "remote": "https://github.com/adamharrison/lite-xl-www.git:aecc866c33b0a954fbd41a5de2842bb97f4ac45e",
       "mod_version": "3"
+    },
+    {
+     "id": "code+",
```
Guldoman, line 1422: I don't know if you can have symbols in the `id`, but you can for sure in the `name` field.
@adamharrison could you confirm?

manifest.json
```diff
@@ -1417,6 +1417,13 @@
...
+     "version": "1.0",
+     "description": "Offers improvements such as highlighted comments and autocomplete for brackets, quotes and more. *([demo](https://s12.gifyu.com/images/Sckre.gif))*",
+     "remote": "https://github.com/chqs-git/code-plus",
```
Guldoman, line 1425: You must specify a commit hash like:
```suggestion
     "remote": "https://github.com/chqs-git/code-plus:b336999b707709df2ddcb865f664623c5beceabb",
```

manifest.json
chqs-git: Thank you, I will make a new commit with the suggested changes. cheers

manifest.json
chqs-git: Out of curiosity can I ask why?

chqs-git: Just understood, thank you.

manifest.json
adamharrison: Confirmed, ids should be `^[a-z0-9\-_]+$` as per [SPEC](https://github.com/lite-xl/lite-xl-plugin-manager/blob/master/SPEC.md).

PR #275, opened 2023-08-08
Dlang: allow underscores for numbers and improve function matching

Tamnac:


Guldoman:
Could you bump the plugin version in the manifest?

PR #277, opened 2023-08-10
wal.lxl: `pywal` integration with Lite XL

ThaCuber:
[yeah, just check the page](https://github.com/ThaCuber/wal.lxl)


ThaCuber:
ok, it should be ready

PR #289, opened 2023-08-27
update wal.lxl

thacuber2a03:
I had recently changed `wal`'s defaults but I completely forgot to reflect it in the repo :sweatsmile:

I also changed the id cuz it's more fun to write `wal` than `wal.lxl`

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1450,8 +1450,8 @@
     },
     {
       "description": "`pywal` integration with Lite XL.",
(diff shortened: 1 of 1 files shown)


Guldoman (changes requested):
> I also changed the id

Then you need to add
```json
"replaces": ["wal_lxl"]
```

thacuber2a03:
done

PR #292, opened 2023-09-13
Link my repository for base16 theme manager for lite-xl

SmileYzn:
Link my repository for base16 theme manager for lite-xl.
Since theme16 was outdated

I had done some work to lite-xl accept base16 themes with yaml extension


adamharrison:
Hey; so the README is auto-generated from the manifest.json; can you modify that instead?

SmileYzn:
There is

adamharrison:
Hey, so if you are going to supply a remote, it has to be in a format that git can wrangle, with an associated commit hash. Something like:

`https://github.com/SmileYzn/style16.git:71b7f873767bbf81efac30ab1ba382b603e3bea3`

In addition, on your repo, you should probably specify a manifest.json there too, that specifies what plugins you're providing and where they are. If you wanted it to be auto-generated from the plugin manager, I think if you put the `base16` folder into `plugins/base16`, it should understand what you mean, and work correctly.

The spec for manifests is here: https://github.com/lite-xl/lite-xl-plugin-manager/blob/master/SPEC.md

SmileYzn:
I have done these changes, i also renamed repository to base16 (And i did not really know why i make as style16 before :rofl:)

I Think it should work now?

Ps. Is needed to put plugin dependencies like widgest and settings on my manifest.json of base16 plugin?

manifest.json
```diff
@@ -1463,6 +1463,13 @@
      "description": "Offers improvements such as highlighted comments and autocomplete for brackets, quotes and more. ([demo](https://s12.gifyu.com/images/Sckre.gif))",
      "remote": "https://github.com/chqs-git/code-plus.git:b336999b707709df2ddcb865f664623c5beceabb",
      "mod_version": 3
+    },
+    {
+      "description": "Theme manager with base16 color schemes for lite-xl",
```
Guldoman, line 1468: Just a nit but
```suggestion
      "description": "Theme manager with base16 color schemes for Lite XL.",
```

manifest.json
```diff
@@ -1463,6 +1463,13 @@
...
+      "version": "0.1",
+      "remote": "https://github.com/SmileYzn/base16.git:latest",
```
Guldoman, line 1470: You can't use `latest` in the main plugins repo, you have to use a commit hash.

Guldoman (commented):
Could you move the entry to where it would fit alphabetically based on the `id`?

Guldoman:
> Ps. Is needed to put plugin dependencies like widgest and settings on my manifest.json of base16 plugin?

I gave a quick look at the plugin, it doesn't seem to *require* them, so I think you can avoid putting them in the manifest.

manifest.json
SmileYzn: Done

SmileYzn:
Settings plugin will make it more dynamic on color change, but in fact should work without it

Thanks for review :pray:

adamharrison:
Just tested it with:

```
lpm run https://github.com/SmileYzn/lite-xl-plugins.git:master base16
```
and
```
lpm run https://github.com/SmileYzn/lite-xl-plugins.git:master --config  'config.plugins.base16.theme = "apathy"'
```

Works for me! Merging.

PR #294, opened 2023-09-14
Update for base16 scheme colors

SmileYzn:
I had to remove table order list to make select box order themes in correctly.


SmileYzn:
Updated scheme list

SmileYzn:
fixed core.error message

PR #297, opened 2023-09-17
Updated workflow to match new lpm changes.

adamharrison:
If we change `lpm` to remove the table action as per https://github.com/lite-xl/lite-xl-plugin-manager/commit/bccca54916aad5b94eed9b619b6af23b36a6aeda, and have a more generic output format, this is what it'd look like.


adamharrison:
To be clear, this should only be merged once bccca54 is released.

Guldoman (commented):
Might be a good idea to put languages before libraries.

adamharrison:
Yup; done.

adamharrison:
OK, new lpm released. So this should be merged ASAP as the current workflow will fail as the `table` command no longer exists.

PR #306, opened 2023-09-25
ipc

zen0bit:
(open in new tab instead of new window)


zen0bit:
I was looking for this funcion, but must ask on discord.
Not obvious from README.md...

Guldoman:
The README is generated from `manifest.json` so that's the one that needs to be updated.

I think something like `Adds inter-process communication support, single-instance mode and tab drag and drop between instances.` would be more explanatory.

PR #308, opened 2023-09-25
`tab_switcher`: Sort entries by last focused and add "hold" mode

Guldoman:
The last entry in `tab-switcher:tab-list{,-current-split}` is the previously focused tab, so that pressing enter is enough to switch to  it.

Also added `tab-switcher:switch-to-last-tab{,-in-current-split}` commands.

Keymap suggestions for the new commands are welcome.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1314,7 +1314,7 @@
     },
     {
       "description": "Switch between open tabs by searching by name",
-      "version": "0.1",
+      "version": "0.2",
       "path": "plugins/tab_switcher.lua",
       "id": "tab_switcher",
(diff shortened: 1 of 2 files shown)


Guldoman:
With the latest commit I added an "hold" mode similar to what other editors do.
This allows one to press `alt+shift+p` and while the modifier keys are still pressed, pressing `p` again will select the next entry. When all the modifier keys are released, the current selection will be accepted.

I'm not sure about the `tab-switcher:tab-list-current-split` and `tab-switcher:switch-to-last-tab-in-current-split` command names anymore, so suggestions are welcome

PR #309, opened 2023-09-28
Add Hybrid Line-numbers option to linenumbers.lua

geetmankar:
Added an option to turn on hybrid line-numbers. When this option is turned on, it overrides the relative-line number option, i.e., if hybrid line numbers is **ON** it doesn't matter if relative line-numbers is **ON** or **OFF**, hybrid line-numbers will take priority.

Additionally, changed the version of linenumbers plugin to `0.2`.

Here is how it looks if the cursor is on line number 15:

image

### DIFF
--- a/plugins/linenumbers.lua
+++ b/plugins/linenumbers.lua
@@ -8,6 +8,7 @@ local command = require "core.command"
 config.plugins.linenumbers = common.merge({
   show = true,
   relative = false,
+  hybrid = false,
   -- The config specification used by the settings gui
   config_spec = {
     name = "Line Numbers",
@@ -24,12 +25,21 @@ config.plugins.linenumbers = common.merge({
       path = "relative",
       type = "toggle",
       default = false
+    },
+    {
+      label = "Hybrid Line Numbers",
+      description = "Display hybrid line numbers starting from active line (Overpowers relative line-numbers).",
+      path = "hybrid",
+      type = "toggle",
+      default = false
     }
   }
 }, config.plugins.linenumbers)
 
 local draw_line_gutter = DocView.draw_line_gutter
 local get_gutter_width = DocView.get_gutter_width
+local isHybridEnabled = false  -- Added flag to track hybrid mode
+
 
 function DocView:draw_line_gutter(line, x, y, width)
   local lh = self:get_line_height()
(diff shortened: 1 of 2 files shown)


plugins/linenumbers.lua
```diff
@@ -24,48 +25,71 @@ config.plugins.linenumbers = common.merge({
...
+      path = "hybrid",
+      type = "toggle",
+      default = false
     }
   }
 }, config.plugins.linenumbers)
 
 local draw_line_gutter = DocView.draw_line_gutter
 local get_gutter_width = DocView.get_gutter_width
+local isHybridEnabled = false  -- Added flag to track hybrid mode
```
Guldoman, line 41: I guess this can be removed, right?

Guldoman (commented):
Could this be simplified with something like:
```Lua
    local local_idx = math.abs(line - l1)
    local alignment = "right"
    local x_offset = style.padding.x
    if config.plugins.linenumbers.hybrid and line == l1 then
        local_idx = line
        alignment = "left"
        x_offset = 0
    end
    common.draw_text(
      self:get_font(),
      color, local_idx, alignment,
      x + x_offset,
      y,
      width, lh
    )
```

plugins/linenumbers.lua
geetmankar: Yes that can be removed

geetmankar: I'll try the simplification

Guldoman: Yeah, that looks fine, thanks!

plugins/linenumbers.lua
```diff
@@ -37,46 +45,47 @@ function DocView:draw_line_gutter(line, x, y, width)
...
 
-    common.draw_text(
-      self:get_font(),
-      color, local_idx, "right",
-      x + style.padding.x,
-      y,
-      width, lh
-    )
-  else
-    return draw_line_gutter(self, line, x, y, width)
```
Guldoman, line 67: You should probably keep this.

geetmankar: Even though this serves the same purpose?:

```lua
.
.
.
elseif not config.plugins.linenumbers.relative then
    local_idx = line
```

Guldoman: It's probably fine to add something like:
```lua
if not config.plugins.linenumbers.relative and not config.plugins.linenumbers.hybrid then
  return draw_line_gutter(self, line, x, y, width)
end
```
near the beginning of the function.

geetmankar: Added and checked by trying on lite-xl

Guldoman:
Looks good, thanks!
I'll wait a few hours before merging this, if anything comes up.

(Actually the last nit I have is with the command names, as they should all start with the plugin name, but it looks like this problem was there before, so whatever...)

geetmankar:
Sweet, it's my first open-source PR (well not really, the first one was literally correcting a syntax error). Really appreciate the editor and plugins.

Hopefully someone can hash out the command names XD.

Guldoman:
> Sweet, it's my first open-source contribution

Awesome first contribution!

> Really appreciate the editor and plugins.

Glad that you're liking it!

> Hopefully someone can hash out the command names XD.

Yeah, no worries, it's not that big of a deal.

PR #312, opened 2023-10-06
Graphical file picker using Zenity

kmafeni04:
With help from @Guldoman, made a plugin that allows for a graphical file picker using the zenity program on linux

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1471,6 +1471,13 @@
       "name": "wal.lxl",
       "mod_version": "3",
       "replaces": [ "wal_lxl" ]
-    }
+    },
+        {
+      "description": "Graphical filepicker using zenity.",
+      "version": "0.1",
+      "path": "plugins/zenity.lua",
+      "id": "zenity",
+      "mod_version": "3"
+    },
   ]
 }
--- a/manifest.json
+++ b/manifest.json
@@ -330,6 +330,13 @@
       "mod_version": "3"
     },
     {
+      "description": "Graphical filepicker using zenity.",
+      "version": "0.1",
+      "path": "plugins/gui_filepicker.lua",
+      "id": "gui_filepicker",
+      "mod_version": "3"
+    },
+    {
       "description": "Dark (or even Mica!) title bar for Lite XL",
       "version": "0.1",
       "remote": "https://github.com/takase1121/lite-xl-immersive-title:cf3a8029ac7154ea53ac819a95d44a6ff102f051",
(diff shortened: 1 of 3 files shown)


manifest.json
```diff
@@ -1471,6 +1471,13 @@
       "name": "wal.lxl",
       "mod_version": "3",
       "replaces": [ "wal_lxl" ]
-    }
+    },
+        {
```
Guldoman, line 1475: Would you mind aligning this?

plugins/zenity.lua
```diff
@@ -0,0 +1,212 @@
...
+			doc:save(abs_path, abs_path)
+			core.log("Saved as " .. '"' .. dv.doc.filename  .. '"')
+		end)
+	end,
+
+	["zenity:save"] = function(dv)
+		local doc = dv.doc
+		if dv.doc.filename then
+			doc:save()
+			core.log("Saved " .. '"' .. dv.doc.filename  .. '"')
```
Guldoman, line 199: I guess you could do
```suggestion
			command.perform("doc:save")
```

plugins/zenity.lua
```diff
@@ -0,0 +1,212 @@
...
+			end
+			local abs_path = table.concat(buffer) or ""
+
+			-- Remove final newline zenity adds
+			abs_path = string.match(abs_path, "^[^\n]+") or ""
+			if #abs_path == 0 then
+				return
+			end
+			doc:save(abs_path, abs_path)
+			core.log("Saved as " .. '"' .. dv.doc.filename  .. '"')
```
Guldoman, line 191:
```suggestion
			core.log('Saved as "%s"', dv.doc.filename)
```

manifest.json
```diff
@@ -1471,6 +1471,13 @@
...
+      "description": "Graphical filepicker using zenity.",
+      "version": "0.1",
+      "path": "plugins/zenity.lua",
+      "id": "zenity",
+      "mod_version": "3"
+    },
```
Guldoman, line 1481:
```suggestion
    }
```
Trailing `,` breaks JSON.

Guldoman (changes requested):
The trailing comma in the manifest is an issue, as it'll break `lpm`.

There is a bunch of repeated code that could be moved into a function, but it's alright for now.

I'm not sold on the plugin ID/filename/command prefixes being just `zenity` tho.

kmafeni04:
> I'm not sold on the plugin ID/filename/command prefixes being just `zenity` tho.

What could be a more suitable name, `gui_save`?

Also will make changes in the morning

Guldoman:
> What could be a more suitable name, `gui_filepicker`?

Yeah, that sounds a bit more descriptive.
I wonder if we should note that `zenity` is only available on linux (and kinda on macOS too).

kmafeni04:
> Yeah, that sounds a bit more descriptive. I wonder if we should note that `zenity` is only available on linux (and kinda on macOS too).

That should be right

PR #316, opened 2023-10-12
fix fatal motiontrail error when mixed with settings

thacuber2a03:
when using settings, clicking on a widget with text capabilities fatally crashes Lite XL
this is a quick hack to fix the issue, but this should be checked on further
(and with "this should be checked" I mean "I should check this")

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1314,7 +1314,7 @@
     },
     {
       "description": "Adds a motion-trail to the caret *([gif](https://user-images.githubusercontent.com/3920290/83256814-085ccb00-a1ab-11ea-9e35-e6633cbed1a9.gif))*",
-      "version": "0.2",
+      "version": "0.3",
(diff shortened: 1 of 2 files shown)


thacuber2a03:
I am pretty sure I bumped motiontrail's version before doing the PR

thacuber2a03:
FUCK I changed the wrong branch

manifest.json
```diff
@@ -1314,7 +1314,7 @@
     },
     {
       "description": "Adds a motion-trail to the caret *([gif](https://user-images.githubusercontent.com/3920290/83256814-085ccb00-a1ab-11ea-9e35-e6633cbed1a9.gif))*",
-      "version": "0.2",
+      "version": "0.3",
```
Guldoman, line 1317: It might be better to do a smaller bump:
```suggestion
      "version": "0.2.1",
```

PR #330, opened 2023-11-02
Update language_rivet.lua

StunxFS:
* Remove keywords not used by Rivet.
* Separate literals from keywords.


plugins/language_rivet.lua
```diff
@@ -1,7 +1,7 @@
 -- mod-version:3
 
 -- Syntax highlighting for the Rivet programming language.
--- by StunxFS :)
+-- By StunxFS.
```
Guldoman, line 4: :(

StunxFS: We need the happy face, right? I forgot to put it in x,D. I'll put it back in another PR :).

PR #331, opened 2023-11-05
Added syntax support for JSON files

c6c534e825f4a3d41ede3e67473187d1:
Looks cleaner than the default syntax `language_js` gives

Example:
2023-11-05-234949_380x215_scrot

### DIFF
--- a/plugins/language_json.lua
+++ b/plugins/language_json.lua
@@ -1,4 +1,4 @@
--- mod-version:3
+-- mod-version:3 priority:90
 
 local syntax = require "core.syntax"
 
--- a/plugins/language_json.lua
+++ b/plugins/language_json.lua
@@ -1,4 +1,4 @@
--- mod-version:3 priority:90
+-- mod-version:3
 
 local syntax = require "core.syntax"
 
@@ -8,8 +8,8 @@ syntax.add {
   comment = nil,
   patterns = {
 
-    { pattern = '\"[a-zA-Z0-9_]*\":', type = "keyword" }, -- key
-    { pattern = { '"', '"', "\\"}, type = "string" }, -- value
+    { pattern = '\"[a-zA-Z0-9_ -]*\":', type = "keyword" }, -- key
+    { pattern = '\"[^\n]*"', type = "keyword" }, -- value
 
     -- numbers
     { pattern = "0x[%da-fA-F]+", type = "number" },
--- a/plugins/language_json.lua
+++ b/plugins/language_json.lua
@@ -4,9 +4,14 @@ local syntax = require "core.syntax"
 
 syntax.add {
   name = "JSON",
-  files = { "%.json$" },
+  files = { "%.json$", "%.cjson$" },
   comment = nil,
   patterns = {
+  
+    -- cjson support
+    { pattern = "//.-\n", type = "comment" },
+    { pattern = { "/%*", "%*/" }, type = "comment" },
+    
     { pattern = '"[^\n]*"%s*:', type = "keyword" }, -- key
     { pattern = '"[^\n]*"', type = "string" }, -- value
     { pattern = "0x[%da-fA-F]+", type = "number" },
--- a/plugins/language_json.lua
+++ b/plugins/language_json.lua
@@ -4,7 +4,7 @@ local syntax = require "core.syntax"
 
 syntax.add {
   name = "JSON",
-  files = { "%.json$", "%.cjson$" },
+  files = { "%.json$", "%.cjson$", "%.jsonc$" },
   comment = "//",
   block_comment = {"/*", "*/"},
   patterns = {
--- a/plugins/language_json.lua
+++ b/plugins/language_json.lua
@@ -13,7 +13,7 @@ syntax.add {
     { pattern = "//.*", type = "comment" },
     { pattern = { "/%*", "%*/" }, type = "comment" },
     
-    { pattern = '"[^\n]-"%s*:', type = "keyword" }, -- key
+    { pattern = '"[^\n]-"()%s*:', type = "normal" }, -- key
     { pattern = '"[^\n]-"', type = "string" }, -- value
     { pattern = "0x[%da-fA-F]+", type = "number" },
     { pattern = "-?%d+[%d%.eE]*",  type = "number" },
--- a/plugins/language_json.lua
+++ b/plugins/language_json.lua
@@ -5,7 +5,8 @@ local syntax = require "core.syntax"
 syntax.add {
   name = "JSON",
   files = { "%.json$", "%.cjson$" },
-  comment = nil,
+  comment = "//",
+  block_comment = {"/*", "*/"},
   patterns = {
   
     -- cjson support
--- a/plugins/language_json.lua
+++ b/plugins/language_json.lua
@@ -1,4 +1,4 @@
--- mod-version:3 priority:90
+-- mod-version:3 priority:110
 
 local syntax = require "core.syntax"
 
@@ -7,8 +7,8 @@ syntax.add {
   files = { "%.json$" },
   comment = nil,
   patterns = {
-    { pattern = '\"[a-zA-Z0-9_ -]*\":', type = "keyword" }, -- key
-    { pattern = '\"[^\n]*"', type = "string" }, -- value
+    { pattern = '"[^\n]*"%s*:', type = "keyword" }, -- key
+    { pattern = '"[^\n]*"', type = "string" }, -- value
     { pattern = "0x[%da-fA-F]+", type = "number" },
     { pattern = "-?%d+[%d%.eE]*",  type = "number" },
     { pattern = "-?%.?%d+", type = "number" },
--- a/plugins/language_json.lua
+++ b/plugins/language_json.lua
@@ -7,16 +7,11 @@ syntax.add {
   files = { "%.json$" },
   comment = nil,
   patterns = {
-
     { pattern = '\"[a-zA-Z0-9_ -]*\":', type = "keyword" }, -- key
-    { pattern = '\"[^\n]*"', type = "keyword" }, -- value
-
-    -- numbers
+    { pattern = '\"[^\n]*"', type = "string" }, -- value
     { pattern = "0x[%da-fA-F]+", type = "number" },
     { pattern = "-?%d+[%d%.eE]*",  type = "number" },
     { pattern = "-?%.?%d+", type = "number" },
-
-    -- literals
     { pattern = "null", type = "literal" },
     { pattern = "true", type = "literal" },
     { pattern = "false", type = "literal" }
--- a/plugins/language_json.lua
+++ b/plugins/language_json.lua
@@ -13,8 +13,8 @@ syntax.add {
     { pattern = "//.*", type = "comment" },
     { pattern = { "/%*", "%*/" }, type = "comment" },
     
-    { pattern = '"[^\n]-"()%s*:', type = "normal" }, -- key
-    { pattern = '"[^\n]-"', type = "string" }, -- value
+    { regex = [["(?:[^"\\]|\\.)*"()\s*:]], type = { "keyword", "normal" } }, -- key
+    { regex = [["(?:[^"\\]|\\.)*"]], type = "string" }, -- value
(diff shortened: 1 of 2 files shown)


plugins/language_json.lua
```diff
@@ -0,0 +1,28 @@
+-- mod-version:3
```
jgmdev, line 1:
```suggestion
-- mod-version:3 priority:90
```

Since language_js.lua already supports .json files the priority of this plugin should be higher than default so it gets loaded and applied before language_js.

c6c534e825f4a3d41ede3e67473187d1: Didn't knew that was possible, cheers!

Maybe the `.json` file extension should be removed from `language_js` aswell, or should it be kept as backup?

jgmdev: I guess it should be kept to provide some json support on default installs but the maintainers will tell.

plugins/language_json.lua
```diff
@@ -0,0 +1,28 @@
...
+    { pattern = { '"', '"', "\\"}, type = "string" }, -- value
+
+    -- numbers
+    { pattern = "0x[%da-fA-F]+", type = "number" },
+    { pattern = "-?%d+[%d%.eE]*",  type = "number" },
+    { pattern = "-?%.?%d+", type = "number" },
+
+    -- literals
+    { pattern = "NaN", type = "literal" },
+    { pattern = "Infinity", type = "literal" },
```
Guldoman, line 22: I wonder if those should be removed, as they're invalid JSON.

plugins/language_json.lua
```diff
@@ -0,0 +1,28 @@
...
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "JSON",
+  files = { "%.json$" },
+  comment = nil,
+  patterns = {
+
+    { pattern = '\"[a-zA-Z0-9_]*\":', type = "keyword" }, -- key (string)
+    { pattern = '[0-9_]*:', type = "keyword" }, -- key (int)
```
Guldoman, line 12: These should probably allow for spaces between the key and `:`.

Also I think you can avoid escaping `"` here.

Are numeric keys allowed in JSON?

plugins/language_json.lua
```diff
@@ -0,0 +1,28 @@
...
+    { pattern = { '"', '"', "\\"}, type = "string" }, -- value
```
Guldoman, line 13: I wonder if this should be adapted to disallow multiline strings.

plugins/language_json.lua
c6c534e825f4a3d41ede3e67473187d1: You're right, numeric keys are not allowed according to the JSON spec. Got them confused with how python handles them.

plugins/language_json.lua
c6c534e825f4a3d41ede3e67473187d1: I only added them because I've seen it in real world examples, but yeah as they're invalid JSON I think it's best to strip them, or let the user re-add them incase it's neccesary

c6c534e825f4a3d41ede3e67473187d1:
Just noticed another issue, keys that contain `-` are highlighted incorrectly:
2023-11-06-135957_348x487_scrot

plugins/language_json.lua
c6c534e825f4a3d41ede3e67473187d1: Coming  back to this after doing some testing, it seems like `language_js` is loaded before `language_json`, but after removing `priority:90` it works fine again

plugins/language_json.lua
c6c534e825f4a3d41ede3e67473187d1: I've got it to ignore strings like:
```
"foo": "

bar",
```

Should it also ignore `"""`, `''` and the like?

plugins/language_json.lua
```diff
@@ -0,0 +1,26 @@
...
+
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "JSON",
+  files = { "%.json$" },
+  comment = nil,
+  patterns = {
+
+    { pattern = '\"[a-zA-Z0-9_]*\":', type = "keyword" }, -- key
```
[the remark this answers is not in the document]
c6c534e825f4a3d41ede3e67473187d1: Yup, added it to the pattern

c6c534e825f4a3d41ede3e67473187d1:
Excuse the commit mess, I am not 100% awake and in a rush aswell

Guldoman:
> Excuse the commit mess, I am not 100% awake and in a rush aswell

Don't worry about it, it'll all get squashed anyways.

plugins/language_json.lua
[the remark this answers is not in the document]
Guldoman: I mean between the end of the string and `:`, like `"blabla"   :`.

plugins/language_json.lua
```diff
@@ -0,0 +1,21 @@
+-- mod-version:3 priority:90
```
Guldoman, line 1: `110` would probably work better I think.

plugins/language_json.lua
```diff
@@ -0,0 +1,21 @@
...
+
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "JSON",
+  files = { "%.json$" },
+  comment = nil,
+  patterns = {
+    { pattern = '\"[a-zA-Z0-9_ -]*\":', type = "keyword" }, -- key
+    { pattern = '\"[^\n]*"', type = "string" }, -- value
```
Guldoman, line 11: I wonder if we could use the second pattern for the initial part of the first, like:
```suggestion
    { pattern = '"[^\n]*"%s*:', type = "keyword" }, -- key
    { pattern = '"[^\n]*"', type = "string" }, -- value
```
So that it would recognize keys of any kind.

plugins/language_json.lua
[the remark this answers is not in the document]
c6c534e825f4a3d41ede3e67473187d1: Ahh, `[ ]*` would do the trick as using `\w` results in getting errors

plugins/language_json.lua
c6c534e825f4a3d41ede3e67473187d1: Yeah, that works

plugins/language_json.lua
Guldoman: In the end the only issue is that it might break when trying to escape strings, but we can live with that for now.

Guldoman (approved):
Looks good. Will wait a bit, then merge.

jgmdev:
Maybe you could also add [.cjson](https://www.webcomand.com/docs/developer_guide/repositories/storage_engines/file_system/cjson/index.html) support which is same as `.json` but adds multiline comments support `/**/` and  single comment lines `//`

c6c534e825f4a3d41ede3e67473187d1:
Yeah, I suppose that shouldn't be too hard to implement

c6c534e825f4a3d41ede3e67473187d1:
Maybe we should eventually add syntax support for cJSON itself in a new file, as it's a bit more than just comments?

Guldoman:
What else does it add?

c6c534e825f4a3d41ede3e67473187d1:
It has a strict format, and a cJSON payload must include objects such as `type` and `version`.
Here's an example:
```
/**
 * Contacts - Some contacts with photos.
 */
{
    "type": "COMAND",
    "version": "1.0",
    "title": "Contacts",
    "description": "Some contacts with photos.",
    "source": "https://comandsystems.com/contacts",
    "time": "2013-10-16T15:39:25+0400",
    "contents": [
        {
            "ID": 13,
            "OID": 2458,
            "UUID": "550e8400-e29b-41d4-a716-446655440013",
            "Type": {"UUID": "550e8400-e29b-41d4-a716-446655440123"},
            "Name": "Mary Smith",
            "EmergencyContact": {"OID": 2460},
            "PhotoFilename": "profile.jpg",
            "Photo": {"type": "url", "value": "files/1.jpg"},
            "Phones": [
                {
                    "ID": 1,
                    "OID": 2478,
                    "UUID": "550e8400-e29b-41d4-a716-446655442323",
                    "Type": {"UUID": "550e8400-e29b-41d4-a716-446655441122"},
                    "Number": "555-555-5555",
                    "Description": "Mobile"
                }
            ]
        },
        {
            "ID": 15,
            "OID": 2460,
            "UUID": "550e8400-e29b-41d4-a716-446655440015",
            "Type": {"UUID": "550e8400-e29b-41d4-a716-446655440123"},
            "Name": "John Smith",
            "EmergencyContact": {"OID": 2458}, // reference to Mary Smith
            "PhotoFilename": "mug.jpg",
            "Photo": {"type": "url", "value": "files/2.jpg"}
        }
    ]
}
```

Guldoman:
Mmm, yeah would make sense to move that in a different language, and highlight those special keys.

Btw I think we're conflating cjson and jsonc. Where the latter is the one that "simply" adds comments (maybe something more, I'd have to check).

jgmdev:
Found some issue while working on a manifest file:

json-issue

plugins/language_json.lua
```diff
@@ -0,0 +1,27 @@
...
+  comment = "//",
+  block_comment = {"/*", "*/"},
+  patterns = {
+  
+    -- cjson support
+    { pattern = "//.*", type = "comment" },
+    { pattern = { "/%*", "%*/" }, type = "comment" },
+    
+    { pattern = '"[^\n]*"%s*:', type = "keyword" }, -- key
+    { pattern = '"[^\n]*"', type = "string" }, -- value
```
Guldoman, line 17:
```suggestion
    { pattern = '"[^\n]-"%s*:', type = "keyword" }, -- key
    { pattern = '"[^\n]-"', type = "string" }, -- value
```

Guldoman: Also, to avoid highlighting the `:`:
```suggestion
    { pattern = '"[^\n]*"%s*():', type = { "keyword", "normal"} }, -- key
    { pattern = '"[^\n]*"', type = "string" }, -- value
```

jgmdev:
Regular expressions to the rescue:

```lua
    { regex = [["(?:[^"\\]|\\.)*"\s*:]], type = "keyword" }, -- key
    { regex = [["(?:[^"\\]|\\.)*"]], type = "string" }, -- value
```

json-with-escape


Regex recipe from: https://stackoverflow.com/a/249937

c6c534e825f4a3d41ede3e67473187d1:
> Regular expressions to the rescue:
> 
> ```lua
>     { regex = [["(?:[^"\\]|\\.)*"\s*:]], type = "keyword" }, -- key
>     { regex = [["(?:[^"\\]|\\.)*"]], type = "string" }, -- value
> ```

Does this also fix the highlighting of `:`?

Guldoman:
> > Regular expressions to the rescue:
> > ```lua
> >     { regex = [["(?:[^"\\]|\\.)*"\s*:]], type = "keyword" }, -- key
> >     { regex = [["(?:[^"\\]|\\.)*"]], type = "string" }, -- value
> > ```
> 
> Does this also fix the highlighting of `:`?

It does not. This will tho:
```lua
    { regex = [["(?:[^"\\]|\\.)*"()\s*:]], type = { "keyword", "normal" } }, -- key
    { regex = [["(?:[^"\\]|\\.)*"]], type = "string" }, -- value
```

c6c534e825f4a3d41ede3e67473187d1:
> ```lua
> { regex = [["(?:[^"\\]|\\.)*"()\s*:]], type = { "keyword", "normal" } }, -- key
>     { regex = [["(?:[^"\\]|\\.)*"]], type = "string" }, -- value
> ```

Just added this, and it seems to be working good! I'll add it in a sec

plugins/language_json.lua
```diff
@@ -0,0 +1,27 @@
+-- mod-version:3 priority:110
+
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "JSON",
+  files = { "%.json$", "%.cjson$" },
```
Guldoman, line 7: I'd also add `.jsonc`.

Guldoman (approved):
Looks good, thank you!
Only one last nit, and after that's fixed I'll wait a bit more and merge.

Guldoman:
Thank you!

PR #333, opened 2023-11-08
fix(language_rivet): fix attribute highlighting

StunxFS:
Use keyword2 instead of comment to highlight.


manifest.json
```diff
@@ -1,4 +1,4 @@
-{
+ya lo{
```
Guldoman, line 1: Something got there by mistake.

StunxFS: Oops 😅, it's fixed now, thanks 🫡

Guldoman:
Looks good, thanks!

Issue #336, opened 2023-11-11
It is possible to add custom extensions to syntax plugins?


felixsanz:
For example `.astro` files works great with the JSX language syntax. It is possible via config (init.lua) to include `.astro` files on this language_jsx.lua ?

```
syntax.add {
  name = "JSX",
  files = { "%.jsx$" },
```
to
```
files = { "%.jsx$", "%.astro$" },
```

Or i'm forced to duplicate the language file and include a modified version?

Guldoman:
You can either use the `force_syntax` plugin (which doesn't save the choice yet), add it to the JSX syntax plugin (and maybe PR that), or you can add this to your user module:
```lua
local core = require "core"
local syntax = require "core.syntax"

core.add_thread(function() -- Use a coroutine to wait until all the plugins are loaded
  for _, s in ipairs(syntax.items) do
    if s.name == "JSX" then
      table.insert(s.files, "%.astro$")
      break
    end
  end
end)
```

felixsanz:
> You can either use the `force_syntax` plugin (which doesn't save the choice yet), add it to the JSX syntax plugin (and maybe PR that), or you can add this to your user module:
> 
> ```lua
> local core = require "core"
> local syntax = require "core.syntax"
> 
> core.add_thread(function() -- Use a coroutine to wait until all the plugins are loaded
>   for _, s in ipairs(syntax.items) do
>     if s.name == "JSX" then
>       table.insert(s.files, "%.astro$")
>       break
>     end
>   end
> end)
> ```

thank for for the snippet. i was trying to do it myself but clearly i wasn't even near. would you accept a PR with astro files being highlighted by JSX plugin?

Guldoman:
> would you accept a PR with astro files being highlighted by JSX plugin?

Yes, thank you!

felixsanz:
@Guldoman I'm an idiot, I wanted to add it to TSX, not JSX.

But anyway... now thinking about it, could be both. It is possible to add the same extension to 2 different language syntax? It depends whether you are using typescript or not inside an astro project (both jsx and tsx are valid).

In fact I tested both, using a typescript astro project, and seems like JSX is even better highlighting.

So can we add the astro extension to language_tsx.lua or lite-xl is not compatible with extensions on multiple grammars?

Also language_tsx.lua is missing the comment fixes you proposed.

Guldoman:
No worries.

[...]

PR #338, opened 2023-11-11
Added astro extension to TSX and fixed block comments

felixsanz:
TSX version of the change already made in JSX.
Fixed some commenting patterns.

#336


Guldoman:
Looks good! Thanks!

PR #340, opened 2023-11-11
update `language_rivet.lua`

StunxFS:
Remove keywords that are no longer supported, improve syntax highlighting for attributes and support syntax highlighting for `defer(MODE)`.

Also support `int` and `uint` types.


plugins/language_rivet.lua
```diff
@@ -30,9 +30,11 @@ syntax.add {
         },
         {pattern = "[A-Z][%w_]*", type = "keyword2"}, -- types
         {pattern = "%@%s?[%a_][%w_]*", type = "literal"}, -- builtin func/var
+        -- `defer` modes
+        {pattern = "defer%s?%(%s?()[%a_][%w_]*()%s?%)", type = {"keyword", "comment"} },
```
Guldoman, line 34: There are 3 "groups", so there should be 3 `type`s.

StunxFS: Oops, that way the word in parentheses `(MODE)` was highlighted

Guldoman:
Looks good, thanks!

PR #342, opened 2023-11-23
Associate ".pm" with Perl icon in Nonicons and Nerdicons

rawleyfowler:


adamharrison:
A fellow perl enthusiast! :open_mouth:

PR #346, opened 2023-12-07
Update language_tal.lua

thacuber2a03:
- no longer using a pattern to match keywords (the language is now a thousand or more characters bigger)
- lambdas are valid jump labels
- DEI and DEO were confused with numbers (I should probably work on that)


plugins/language_tal.lua
```diff
@@ -1,29 +1,28 @@
 --mod-version:3
-local syntax = require 'core.syntax'
 
 local label, sublabel = "function", "keyword2"
 
-syntax.add {
+require 'core.syntax'.add {
```
Guldoman, line 5: Why?

plugins/language_tal.lua
```diff
@@ -1,29 +1,28 @@
...
-		{ pattern = "%.%S+()/%S*",      type = { label, sublabel } },
-		{ pattern = "%.%S+",            type = label      },
-		{ pattern = "|%x+",             type = "string"   },
-		{ pattern = "[.,;_=-]%S+",      type = sublabel   },
-		{ pattern = "%$%d+",            type = "number"   },
-		{ pattern = "#?%x%x%x%x%f[%X]", type = "number"   },
-		{ pattern = "#?%x%x%f[%X]",     type = "number"   },
-		{ pattern = "[!?]()[^%[%]{}%s]+", type = { "operator", "function" } },
-		{ pattern = "[^%[%]{}%s]+",     type = "function" },
+		{ pattern = { "%f[%(]%(%s", "%)" }, type = "comment"  },
```
Guldoman, line 10: What's the reason for this change?

thacuber2a03: comment delimiters must have at least a space between them and any other character to be recognized as comments, that's just an easier way to notice when you forget to put space in at least the left side, because I don't quite know how to catch the right side too :/

plugins/language_tal.lua
thacuber2a03: core.syntax is only used that one time

Guldoman: Yeah, like basically every other language syntax plugin. No reason to make this less legible.

plugins/language_tal.lua
Guldoman: Wait, could you make some examples of valid and invalid comments?

plugins/language_tal.lua
thacuber2a03: this is arguably just as readable, but alright then

plugins/language_tal.lua
thacuber2a03: - `(invalid)`, `(invalid )`, `( invalid)`
- `( valid )`, `(     valid   )`, `( valid        )`
at least according to uxnasm and Drifblim

Guldoman: No requirements about what comes before `(` and after `)`?

thacuber2a03: the extra spaces won't show up :(

Guldoman: Looks fine here:
image

Guldoman: If there aren't requirements before/after the comment, wouldn't `{ "%(%s", "%s%)" }` work fine?

thacuber2a03: > No requirements about what comes before `(` and after `)`?

there can't be any character before or after the `(`, and the right hand side only closes when the `)` is alone

as in, `( test) some more stuff` will eat everything after the `)`

Guldoman: I guess you can either make this a regex, or split it into multiple patterns that match the various cases.

PR #351, opened 2024-01-03
Update manifest.json for lsp_zig

ArtemisX64:


Guldoman (changes requested):
There is no `0.11.1` in the remote repo, so we can't merge this.

You also need to update the commit hash in the `remote` entry to match the one in the remote repo that contains the version you want.

Issue #236, opened 2023-04-25
[Request] Add Typst syntax highlighting 


knedl1k:
References:
VSCode/Codium extension: https://github.com/nvarner/typst-lsp

language: https://github.com/typst/typst

Guldoman:
Added in #361!

PR #361, opened 2024-02-08
Add support for Typst synatx

Gaspartcho:
Hello, this is the first time I'm doing this (I had to learn everything on the fly), so the syntax itself is probably very broken.

This file includes the bare-bones of the support of Typst syntax, with an emphasis on the math part (as I use it mostly to write my math homework). Hence, everything concerning scripting here might be very broken (not for the lack of trying though, I just find the syntax confusing to implement)...

This PR should also resolve Issue #236 (or at least begin to).

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -148,6 +148,7 @@
         "language_toml": {},
         "language_ts": {},
         "language_tsx": {},
+        "language_typst": {},
         "language_umka": {},
         "language_v": {},
         "language_wren": {},
@@ -1232,6 +1233,14 @@
       "version": "0.2",
       "path": "plugins/language_tsx.lua",
       "id": "language_tsx",
+      "mod_version": "3",
+      "tags": ["language"]
+    },
+    {
+      "description": "Syntax for [Typst](https://typst.app/) markup language",
+      "version": "0.1",
+      "path": "plugins/language_typst.lua",
+      "id": "language_typst",
       "mod_version": "3",
       "tags": ["language"]
     },
--- a/manifest.json
+++ b/manifest.json
@@ -61,7 +61,7 @@
     {
       "id": "meta_languages",
       "type": "meta",
-      "version": "0.1.5",
+      "version": "0.1.6",
       "mod_version": "3",
       "description": "A metapackage containing all publically accessible language syntaxes.",
       "dependencies": {
--- a/plugins/language_typst.lua
+++ b/plugins/language_typst.lua
@@ -136,7 +136,7 @@ syntax.add {
 
    --     { pattern = {"link"},                    type = "typst_underline"}, -- add underline text later
 
-        { pattern = "^ *=+ ().+%f[\n]",              type ={"operator", "literal"}}, -- Supposed to be bold
+        { pattern = "^%s*=+ ().+%f[\n]",             type ={"operator", "literal"}}, -- Supposed to be bold
 
     -- Finish later
         { pattern = {"%*_", "_%*[%s,%.]"},         type = "keyword2"}, -- Bold Italic
--- a/plugins/language_typst.lua
+++ b/plugins/language_typst.lua
@@ -2,9 +2,9 @@
 local syntax = require "core.syntax"
 local style = require "core.style"
 
--- This syntax is probably very borken, so feel free to improve it.
 
 --region add other styles
+--[[
 local new_styles = {"bold", "italic", "bold_italic", "underline"}
 
 local code_font_size = style.code_font:get_size()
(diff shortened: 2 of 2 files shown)


manifest.json
```diff
@@ -148,6 +148,7 @@
         "language_toml": {},
         "language_ts": {},
         "language_tsx": {},
+        "language_typst": {},
```
Guldoman, line 151: Would you mind bumping the version of `meta_languages` too?

plugins/language_typst.lua
```diff
@@ -0,0 +1,196 @@
...
+    attributes["bold"] = true
+    attributes["italic"] = true
+  end
+  -- yes way to copy user custom font with additional attributes :)
+  style.syntax_fonts["typst_"..attr] = renderer.font.load(
+    code_font_path,
+    code_font_size,
+    attributes
+  )
+end
```
Guldoman, line 31: I'm sorry for this, but could you avoid adding other token types?
The code as it is now will probably work, but will be broken for people that customize their fonts (even at runtime).

We intend to add better support for this use-case in the future, but it's not there yet.

Guldoman (commented):
Instead of doing `syntax.add` for the subsyntaxes that you don't want to be "public", you can just keep the table in a local variable, and use it with `syntax = subsyntax_table`.
This way you don't even need to add `name` and `files`. Also `comment` and `block_comment` would be inherited from the "main" syntax.

plugins/language_typst.lua
Gaspartcho: Sure, but is there another way I can make the text bold, or italic (or even underlined, even though I could not get that to work in the first place) ?

Guldoman: Sadly not yet. But it is planned for the future.

plugins/language_typst.lua
```diff
@@ -0,0 +1,187 @@
+-- mod-version:3
+local syntax = require "core.syntax"
+local style = require "core.style"
```
Guldoman, line 3: Can be removed.

Guldoman (commented):
There are still a bunch of custom token types.

Other than that, this looks fine to me!

plugins/language_typst.lua
```diff
@@ -0,0 +1,157 @@
...
+
+   --     { pattern = {"link"},                    type = "typst_underline"}, -- add underline text later
+
+        { pattern = "^=+ ().+%f[\n]",              type ={"operator", "normal"}}, -- Supposed to be bold
+
+    -- Finish later
+    --    { pattern = {"%*_", "_%*[%s,%.]"},     type = "bold_italic"},
+    --    { pattern = {"_%*", "%*_[%s,%.]"},     type = "bold_italic"},
+    --    { pattern = {"_", "_[%s,%.]"},         type = "italic"},
+    --    { pattern = {"%*[^%/]", "%*[%s,%.]"},  type = "bold"},
```
Guldoman, line 146: Instead of leaving them out, what do you think about at least highlighting them?
Feel free to use any token type that looks fine.

plugins/language_typst.lua
```diff
@@ -0,0 +1,157 @@
...
+
+        { pattern = { '"', '"', '\\' },            type = "string"  },
+
+
+        { pattern = "//.*",                        type = "comment" },
+        { pattern = { "/%*", "%*/" },              type = "comment" },
+
+   --     { pattern = {"link"},                    type = "typst_underline"}, -- add underline text later
+
+        { pattern = "^=+ ().+%f[\n]",              type ={"operator", "literal"}}, -- Supposed to be bold
```
Guldoman, line 140: Should this allow whitespace before `=`?

Gaspartcho: Actually, according to the [documentation](https://typst.app/docs/reference/model/heading/), there is supposed to be no space before the `=`, and one after.

Guldoman: I was looking at the first example here https://typst.app/docs/tutorial/formatting/#set-rules.

Gaspartcho: Right.

This should be fixed now.

plugins/language_typst.lua
```diff
@@ -0,0 +1,156 @@
...
+
+        { pattern = { '"', '"', '\\' },            type = "string"  },
+
+
+        { pattern = "//.*",                        type = "comment" },
+        { pattern = { "/%*", "%*/" },              type = "comment" },
+
+   --     { pattern = {"link"},                    type = "typst_underline"}, -- add underline text later
+
+        { pattern = "^ *=+ ().+%f[\n]",              type ={"operator", "literal"}}, -- Supposed to be bold
```
Guldoman, line 139: Might be a better idea to allow any whitespace (so tabs too):
```suggestion
        { pattern = "^%s*=+ ().+%f[\n]",             type ={"operator", "literal"}}, -- Supposed to be bold
```

Gaspartcho: Sure

Guldoman:
LGTM, thanks!

PR #363, opened 2024-02-11
Embedded JavaScript templating syntax support

kemzops:
**language_ejs.lua**, fork of language_html.lua to support **`.ejs` files**.

### DIFF
--- a/plugins/language_ejs.lua
+++ b/plugins/language_ejs.lua
@@ -1,0 +1,43 @@
+-- mod-version:3
+
+-- Embedded JavaScript templating
+-- provides .ejs syntax support (fork of language_html.lua).
+
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "EJS",
+  files = { "%.ejs$" },
+  block_comment = { "<!--", "-->" },
+  patterns = {
+    {
+      pattern = {
+        "<%s*[sS][cC][rR][iI][pP][tT]%f[%s>].->",
+        "<%s*/%s*[sS][cC][rR][iI][pP][tT]%s*>"
+      },
+      syntax = ".js",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -267,6 +267,7 @@ but only with a `url` must provide a `checksum` that matches the existing plugin
...
 | [`language_yaml`](plugins/language_yaml.lua?raw=1)                                                               | Syntax for [YAML](https://yaml.org/) serialization language                                                                                                                         |
 | [`language_zig`](plugins/language_zig.lua?raw=1)                                                                 | Syntax for the [Zig](https://ziglang.org/) programming language                                                                                                                     |
+| [`language_ejs`](plugins/language_ejs.lua?raw=1)                                                                 | Syntax for the [EJS](https://ejs.co/)  javascript template engine                                                                                                                     |
```
Guldoman, line 270: No need for this, as this file is generated from `manifest.json`, which should be edited instead.

kemzops: Oh sorry i didn't notice that i thought its automatic, done now i guess @Guldoman .

manifest.json
```diff
@@ -153,7 +153,8 @@
         "language_v": {},
         "language_wren": {},
         "language_yaml": {},
-        "language_zig": {}
+        "language_zig": {},
+        "language_ejs": {}
```
Guldoman, line 157: Could you sort this alphabetically?
Also bump the version of `meta_languages`.

Guldoman (commented):
What do you think about adding a pattern that links `<%` and `%>` to the javascript language, as this seems to be the point of this templating language.

Guldoman:
Looks good! Thanks!

PR #366, opened 2024-02-18
lite-debugger improvements

takase1121:
This PR adds a few improvements to the debugger:

1. hooks are unset if no breakpoints are set.
2. added support for full path breakpoints (e.g. `~/.config/lite-xl/init.lua:blabla` instead of only `init.lua`, which could literally be any module)
3. added a command to list all breakpoints
4. allow spaces before and after a command

cc @Jan200101 since iirc you're the one who added this plugin initially

### DIFF
--- a/plugins/lite-debugger.lua
+++ b/plugins/lite-debugger.lua
@@ -520,6 +520,7 @@ local commands = {
 }
 
 local function match_command(line)
+   line = line:gsub("^%s+", ""):gsub("%s*$", "")
    for pat, func in pairs(commands) do
       -- Return the matching command and capture argument.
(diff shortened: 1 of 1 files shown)


plugins/lite-debugger.lua
```diff
@@ -520,6 +555,7 @@ local commands = {
 }
 
 local function match_command(line)
+   line = line:gsub("^%s+", ""):gsub("%s*$", "")
```
Guldoman, line 558:
```suggestion
   line = line:match("^%s*(.-)%s*$")
```

Guldoman (commented):
Could you bump the version in the manifest?

PR #369, opened 2024-02-24
Remove `linter` plugin

Guldoman:
The plugin doesn't even support Lite XL.
Set `lintplus` as replacement.

@adamharrison does the `replaces` usage make sense? Do we need to keep `linter` around for it to be effective?


adamharrison:
This should work fine, and if it doesn't, I'll fix it.

adamharrison:
it doesn't work; but i'll file an issue in lpm

PR #372, opened 2024-02-28
Add language_djot

Tamnac:
It's a separate repo because it will probably change rapidly as I add in support for previewing.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -691,6 +691,14 @@
       "mod_version": "3",
       "tags": ["language"]
     },
+     {
+      "description": "Syntax for the [Djot](https://djot.net/) markup languiage",
+      "version": "0.1",
+      "path": "https://github.com/Tamnac/lite-xl-djot.git",
+      "id": "language_djot",
+      "mod_version": "3",
+      "tags": ["language"]
+    },
     {
       "description": "Syntax for [FreeFEM++](http://freefem.org) programming language",
       "version": "0.1",
--- a/manifest.json
+++ b/manifest.json
@@ -694,7 +694,7 @@
      {
       "description": "Syntax for the [Djot](https://djot.net/) markup languiage",
       "version": "0.1",
-      "path": "https://github.com/Tamnac/lite-xl-djot.git",
+      "remote": "https://github.com/Tamnac/lite-xl-djot:46e643913dd075c5a12f08e8657f99686948d2cf",
       "id": "language_djot",
       "mod_version": "3",
       "tags": ["language"]
--- a/manifest.json
+++ b/manifest.json
@@ -691,7 +691,7 @@
       "mod_version": "3",
       "tags": ["language"]
     },
-     {
+    {
       "description": "Syntax for the [Djot](https://djot.net/) markup language",
       "version": "0.1",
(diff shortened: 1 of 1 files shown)


manifest.json
```diff
@@ -690,6 +690,14 @@
       "id": "language_diff",
       "mod_version": "3",
       "tags": ["language"]
+    },
+     {
+      "description": "Syntax for the [Djot](https://djot.net/) markup languiage",
+      "version": "0.1",
+      "path": "https://github.com/Tamnac/lite-xl-djot.git",
```
Guldoman, line 697:
```suggestion
      "remote": "https://github.com/Tamnac/lite-xl-djot:46e643913dd075c5a12f08e8657f99686948d2cf",
```

Guldoman:
It looks like this:
https://github.com/Tamnac/lite-xl-djot/blob/46e643913dd075c5a12f08e8657f99686948d2cf/language_djot.lua#L6
is breaking the generation of the manifest and should be removed, also because there is no `djot` plugin yet.

Ideally a `language_*` plugin would only implement the tokenizer syntax, leaving more complex things to other plugins.

Guldoman:
> do you think I should add all the other functionality into a separate plugin?

Yeah, better to keep the `language_*` plugins relatively simple.

You could create "bundle" meta-plugins that install multiple plugins at once, so you could have like `bundle_djolt` and it would install `language_djolt`, `lsp_djolt`, `djolt_preview` or whatever.

manifest.json
```diff
@@ -690,6 +690,14 @@
...
       "mod_version": "3",
       "tags": ["language"]
+    },
+     {
+      "description": "Syntax for the [Djot](https://djot.net/) markup languiage",
+      "version": "0.1",
+      "remote": "https://github.com/Tamnac/lite-xl-djot:46e643913dd075c5a12f08e8657f99686948d2cf",
+      "id": "language_djot",
+      "mod_version": "3",
+      "tags": ["language"]
```
Guldoman, line 700:
```suggestion
    {
      "description": "Syntax for the [Djot](https://djot.net/) markup languiage",
      "version": "0.1",
      "remote": "https://github.com/Tamnac/lite-xl-djot:e92ad996aba56c3363ef83a2dd8d1026be56c730",
      "id": "language_djot",
      "mod_version": "3",
      "tags": ["language"]
```

Guldoman:
> Is there a reason you're referencing a specific commit and not just the repo?

Various reasons:
* Keep track of the version, without having to download the linked repo.
* Avoid that malicious entities alter the published plugin in dangerous ways.
* Allows to link to a known working version of the plugin, so that breaking commits don't end up in users' systems.

> Ideally people would just get the latest version of the plugin.

Normally you just push the updated version as a PR here, and it should get quickly accepted.

The commit linking is a policy of this repo. Users can manually add other repos, without needing to link to a specific commit.
For example users could add `https://github.com/Tamnac/lite-xl-djot:master` as a repo and pull the plugin from there.

> Also I'm not familiar with the format of the link, but it doesn't seem to work

It's not a normal link, it's parsed by `lpm` to detect the commit hash to pull.

Guldoman:
Thank you!

PR #373, opened 2024-02-29
`icons` mjs, cjs, pug, ejs

kemzops:
- Added `.mjs` `.cjs` for both nerdicons and nonicons.
- Added `.ejs` `.pug` for nerdicons.

After checking [nvim-nonicons](https://github.com/yamatsum/nvim-nonicons/blob/main/lua/nvim-web-devicons/override.lua) nonicons can support some other icons (Ex: `.ejs`) **BUT** using the old way not the new one (the mapping).

And i noticed a lot of icons that can be supported is there any better way of handling icons ?


Guldoman:
Could you bump the plugins' versions?

PR #380, opened 2024-03-09
Fixing ghmarkdown adding the ability to use a GitHub token

walkero-gr:
Based on changes GitHub did to the `https://api.github.com/markdown/raw` endpoint, which is the one that is used with this plugin, the rendering is failing with the following error message.

```
{ "message": "Invalid request media type (expecting 'text/plain')", "documentation_url": "https://docs.github.com/rest/markdown/markdown#render-a-markdown-document-in-raw-mode" } 
```

The error message is misguiding, though, because the actual problem is the lack of a token user. This PR adds a configuration parameter to the plugin, for the user to easily provide their own token.


Guldoman (commented):
Would you mind bumping the version in `manifest.json` (and updating the description there too).

walkero-gr:
@Guldoman `manifest.json`   file update. Sorry I didn't do it earlier, but I was not aware it was needed.

Guldoman:
> Sorry I didn't do it earlier, but I was not aware it was needed.

No worries!

Looks good, thanks!

PR #381, opened 2024-03-09
Use original CWD when opening files in the ipc plugin

DUOLabs333:
Fixes an issue where if you're using the ipc plugin, you can only open a file from the terminal if you are in the same directory as it.

### DIFF
--- a/plugins/ipc.lua
+++ b/plugins/ipc.lua
@@ -876,6 +876,27 @@ system.get_time = function()
     local primary_instance = ipc:get_primary_instance()
     if primary_instance and ARGS[2] then
       local open_directory = false
+
+      local path_sep=package.config:sub(1,1)
+      local pwd=""
+
+      if path_sep=='\\' then
+        pwd="cd"
+      else
+        pwd="pwd"
+      end
+      
+      local new_dir=os.getenv(string.upper(pwd)) -- This is preserved even though the actual cwd may change
+      local old_dir=io.popen(pwd):read()
+      local should_change_dir=false
+      if new_dir and old_dir then
+        should_change_dir=true -- Hacky. Should be fixed properly by the main program not changing the cwd all of the time
+      end
+
+      if should_change_dir then
+        system.chdir(new_dir)
+      end
+      
       for i=2, #ARGS do
         local path = system.absolute_path(ARGS[i])
 
@@ -900,6 +921,11 @@ system.get_time = function()
           end
         end
       end
+
+      if should_change_dir then
+        system.chdir(old_dir)
+      end
+      
       ipc:wait_for_messages()
       if not open_directory then
         os.exit()


adamharrison:
So wait; perhaps I'm just not seeing it; but could we not simply do something like 

```
local dir = os.getenv(PLATFORM == "windows" and "CD" or "PWD")
local path = dir .. PATHSEP .. ARGS[i]
```

Instead of changing directories? I'm not sure we even need to use `system.absolute_path` here if we do something like this.

DUOLabs333:
I guess, I just want to avoid missing some edge case(s) that `absolute_path` handles.

DUOLabs333:
Also, that wouldn't work for something like `~/Repos` and `~/Repos/wgpu/Cargo.toml`. I'm not even sure what would happen if you just concatenated them and tried to open it.

adamharrison:
Oh, good point. Ideally though, we just minimize the amount of global state we have to deal with. How about something like this, then? 

```
local dir = os.getenv(PLATFORM == "Windows" and "CD" or "PWD")
local path = common.home_expand(ARGS[i])
if not common.is_absolute_path(path) then
  path = dir .. PATHSEP .. path
end
path = system.absolute_path(path)
```

DUOLabs333:
That'll probably work.

Guldoman:
> Also, that wouldn't work for something like `~/Repos` and `~/Repos/wgpu/Cargo.toml`. I'm not even sure what would happen if you just concatenated them and tried to open it.

You should never be getting `~` from the command line anyways. It will be expanded by the shell when calling Lite XL.

DUOLabs333:
Oh, I just used `~` as a shorthand so I wouldn't have the write the full thing here.

Guldoman:
Then an absolute path check should be enough, right?

DUOLabs333:
Yes.

Guldoman:
Could you try with something like:
1. Get "original" `cwd` via `local original_cwd = system.absolute_path(".")` on plugin load, so outside any functions.
2. Add a function like:
   ```lua
   local function get_absolute_path(filename)
     if common.is_absolute_path(filename) then
       return common.normalize_path(filename)
     else
       return original_cwd .. PATHSEP .. common.normalize_path(filename)
     end
   end
   ```
3. Replace `local path = system.absolute_path(ARGS[i])` with the above function `local path = get_absolute_path(ARGS[i])`.

DUOLabs333:
1. The problem is that by the time the plugin is loaded, the CWD has changed.

Guldoman:
Ah yeah, it's the user module that gets loaded before, alright.

Guldoman:
Still, instead of the weird double cwd change, you could still use PWD from the environment variable in my proposed function instead of `original_cwd`.

PR #383, opened 2024-03-13
Update indentguide: Set the curent guide line highlight as optional

Gaspartcho:
Before:
image

After:
image


kemzops:
LGTM! but if you just bumped the version to `0.2` instead of `0.1.1` would be better.

Gaspartcho:
Done!

Guldoman:
Looks good, thank you!

PR #389, opened 2024-03-17
Update typst syntax

Gaspartcho:
I added a lot of things in this one, but it should work far better than before, especially on the scripting part.

Reference: https://typst.app/docs/reference/syntax/

### DIFF
--- a/plugins/language_typst.lua
+++ b/plugins/language_typst.lua
@@ -180,7 +180,7 @@ local typst_syntax = {
 			syntax  = typst_script
 		},
 
-		{ pattern = { "#", "%s"},               type = "literal", syntax = typst_script },
+		{ pattern = { "#", "%f[%s']"},          type = "literal", syntax = typst_script },
 		{ pattern = { "%$", "%$", "\\" },       type = "literal", syntax = typst_math   },
 
 		{ pattern = { "```js"  , "```", "\\" }, type = "string", syntax = ".js"         },
--- a/plugins/language_typst.lua
+++ b/plugins/language_typst.lua
@@ -104,7 +104,7 @@ local typst_script = {
 		{ pattern = { '"', '"', '\\' },            type = "string", syntax = typst_string },
 		{ pattern = {"%[", "%]"},                  type = "normal", syntax = ".typ"       },
 		{ pattern = "[%a_][%w_%-]*%f[(]",          type = "function"                      },
-		{ pattern = "[%+%-=/%*%^%%<>~|&#_^\\%%]",  type = "operator"                      },
+		{ pattern = "[%+%-=/%*%^%%<>~|&#_^\\]",    type = "operator"                      },
 
 		{ pattern = "0x[%dabcdef]+()%a*",          type = {"number", "keyword2"}          },
 		{ pattern = "0b[01]+()%a*",                type = {"number", "keyword2"}          },
@@ -153,6 +153,7 @@ local typst_syntax = {
 	block_comment = { "/*", "*/" },
 
 	patterns = {
+		{ pattern = { "%[", "%]" }, syntax = ".typ" },
 		{
 			pattern = { "#()import", "\n"    },
 			type    = { "literal", "keyword" },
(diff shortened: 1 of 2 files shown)


plugins/language_typst.lua
```diff
@@ -4,153 +4,213 @@ local syntax = require "core.syntax"
...
-        { pattern = {"%*[^%/]", "%*[%s,%.]"},      type = "literal"}, -- Bold
+local typst_script = {
 
-        { pattern = "[%+%-\\]",    type = "operator" },
+	patterns = {
+
+		{ pattern = { "/%*", "%*/" },              type = "comment"                       },
+		{ pattern = "//.*",                        type = "comment"                       },
+
+		{ pattern = { '"', '"', '\\' },            type = "string", syntax = typst_string },
```
Guldoman, line 104: What issue does this solve?

plugins/language_typst.lua
```diff
@@ -4,153 +4,213 @@ local syntax = require "core.syntax"
...
+			type    = { "literal", "keyword" },
+			syntax  = typst_script
+		},
+
+		{ pattern = { "#", "%s"},               type = "literal", syntax = typst_script },
+		{ pattern = { "%$", "%$", "\\" },       type = "literal", syntax = typst_math   },
+
+		{ pattern = { "```js", "```", "\\" },   type = "string", syntax = ".js"         },
+		{ pattern = { "```py", "```", "\\" },   type = "string", syntax = ".py"         },
+		-- Possibility to add more syntax later
```
Guldoman, line 187: You can freely add any you want here, even if they're not shipped by default.

plugins/language_typst.lua
```diff
@@ -4,153 +4,213 @@ local syntax = require "core.syntax"
...
+		{ pattern = "\\[nrt]",                  type = "string"                         },
+		{ pattern = "\\u%b{}",                  type = "string"                         },
+		{ pattern = "\\.",                      type = "normal"                         },
+
+	},
+
+	symbols = {}
+}
 
+table.insert(typst_syntax.patterns, 1, { pattern = { "%[", "%]" }, syntax = typst_syntax })
```
Guldoman, line 212: In this case you could probably get away with adding it at the top, in the actual table, and just use `syntax = ".typ"`.

plugins/language_typst.lua
```diff
@@ -4,153 +4,213 @@ local syntax = require "core.syntax"
...
+		{ pattern = "0x[%dabcdef]+",               type = "number"   },
+		{ pattern = "0b[01]+",                     type = "number"   },
+		{ pattern = "0o[01234567]+",               type = "number"   },
+		{ pattern = "%d+[.e]%d+",                  type = "number"   },
+		{ pattern = "%d",                          type = "number"   },
+
+		{ pattern = "&",                           type = "literal"  },
+		{ pattern = "!=",                          type = "operator" },
+		{ pattern = "[%+%-=/%*%^%%<>~|#_^\\]",     type = "operator" },
+		{ pattern = "%a%a+",                       type = "symbol"   },
```
Guldoman, line 27: Is this to force 2+ characters?

plugins/language_typst.lua
```diff
@@ -4,153 +4,213 @@ local syntax = require "core.syntax"
...
+		-- Possibility to add more syntax later
+		{ pattern = { "`", "`", "\\" },         type = "string"                         },
+
+		{ pattern = "//.*",                     type = "comment"                        },
+		{ pattern = { "/%*", "%*/" },           type = "comment"                        },
+
+		--{ pattern = {"link"},                      type = "typst_underline"                }, -- add underline text later
+
+		-- Finish later
+		{ pattern = "^%s*=+%s.+%f[\n]",         type = "keyword"                        }, -- Bold
```
Guldoman, line 196: I don't think you need to use `%f[\n]`:
```suggestion
		{ pattern = "^%s*=+%s.*",               type = "keyword"                        }, -- Bold
```

plugins/language_typst.lua
Gaspartcho: Yes.

plugins/language_typst.lua
Gaspartcho: There was a bug where the string highlighting did not work properly (probably because the tokenizer prioritizes subsyntaxes over classic start-end patterns). So in order to fix this issue, I had to create this extra sub-syntax.

Guldoman: Could you give me an example?
Just to check if it's something we need to fix, to understand the problem better.

Gaspartcho: Without the extra syntax:
image

With the extra syntax:
image

Guldoman: The issue is caused by
https://github.com/lite-xl/lite-xl-plugins/blob/553b178d929374f4bd2aba50851edbdbbd8986a9/plugins/language_typst.lua#L183

which is stopping at the first space.
Is that intended? Or should it go ahead until the newline?

Gaspartcho: It is intended, you are supposed to do something like that:
```
#let name = "Typst"
This is #name's documentation.
It explains #name.
```

Guldoman: Then it might be a good idea to provide both versions.
One for full line match, and another for "up to the fist space".
For the full line one, you should probably use `^` at the start.

Gaspartcho: This is sort of what I did: I noticed that when it needed a full line match, it was when using keywords (as `let`, `set`, `import` etc...)
In all other cases, it stops at the first space encountered.
This works for most of the cases except for the case where you use directly a string after the `#`, which is why I had to add this extra sub-syntax.

And also, if I used `^` in order to detect a full line capture, something like this wouldn't work:

```
#let name = "Typst"
#name is great!
```

Guldoman: Then I suggest using:
```lua
local typst_string = {

	patterns = {
		{ pattern = "\\.",   type = "string" },
		{ pattern = '[^"\\]+', type = "string" }
	},

	symbols = {}
}
```
This way it matches as much as possible at once, reducing the number of times it needs to run.

Gaspartcho: Thanks! I've added it (and a similar thing for the python syntax).

PR #392, opened 2024-03-26
add font preview plugin

thacuber2a03:
lets you quickly preview a font in the editor for whatever reason

### DIFF
--- a/plugins/fontpreview.lua
+++ b/plugins/fontpreview.lua
@@ -20,6 +20,8 @@ function FontView:get_h_scrollable_size()
 	return self.fonts[#self.fonts]:get_width(font_text) + style.padding.x
 end
 
+function FontView:get_scrollable_size() return 0 end
+
 function FontView:get_name()
 	return "Font: " .. self.path
 end
--- a/plugins/fontpreview.lua
+++ b/plugins/fontpreview.lua
@@ -42,17 +42,22 @@ function FontView:draw()
 	self:draw_scrollbar()
 end
 
+local supported_types = { ".ttf", ".otf" }
+
 local RootView = require 'core.rootview'
 local open_doc = RootView.open_doc
 function RootView:open_doc(doc)
 	local path = doc.filename or doc.abs_filename or ""
-	if path:find(".ttf") then
-		local node = self:get_active_node_default()
-		local view = FontView(path)
-		node:add_view(view)
-		self.root_node:update_layout()
-		return view
-	else
-		open_doc(self, doc)
+
+	for _, v in ipairs(supported_types) do
+		if path:find(v) then
+			local node = self:get_active_node_default()
+			local view = FontView(path)
+			node:add_view(view)
+			self.root_node:update_layout()
+			return view
+		end
 	end
+
+	return open_doc(self, doc)
 end
--- a/plugins/fontpreview.lua
+++ b/plugins/fontpreview.lua
@@ -1,0 +1,63 @@
+--mod-version:3 --priority:110
+
+local common = require 'core.common'
+local style = require 'core.style'
+local View = require 'core.view'
+
+local FontView = View:extend()
+
+local font_text = "The quick brown fox jumps over the lazy dog."
+
+function FontView:new(path)
+	FontView.super.new(self)
+	self.path = path
+	self.fonts = {}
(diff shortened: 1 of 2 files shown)


plugins/fontpreview.lua
```diff
@@ -0,0 +1,63 @@
...
+
+local FontView = View:extend()
+
+local font_text = "The quick brown fox jumps over the lazy dog."
+
+function FontView:new(path)
+	FontView.super.new(self)
+	self.path = path
+	self.fonts = {}
+	for i=1, 8 do self.fonts[i] = renderer.font.load(path, 12+i*7) end
```
Guldoman, line 15: Might be a good idea to scale the font size by `SCALE`.

plugins/fontpreview.lua
```diff
@@ -0,0 +1,63 @@
...
+function FontView:get_h_scrollable_size()
+	return self.fonts[#self.fonts]:get_width(font_text) + style.padding.x
+end
+
+function FontView:get_scrollable_size()
+	return math.huge
+end
+
+function FontView:get_name()
+	return "Font: " .. self.path
```
Guldoman, line 28:
```suggestion
	return "Font: " .. common.basename(self.path)
```
might be a good idea.

plugins/fontpreview.lua
```diff
@@ -0,0 +1,63 @@
...
+	end
+
+	self:draw_scrollbar()
+end
+
+local RootView = require 'core.rootview'
+local open_doc = RootView.open_doc
+function RootView:open_doc(doc)
+	local path = doc.filename or doc.abs_filename or ""
+	if path:find(".ttf") then
```
Guldoman, line 54: You should check other extensions too.
For example `.otf` should be supported.

Guldoman (commented):
When scrolling vertically, text moves weirdly.

I think you can return a sensible value from `FontView:get_scrollable_size`, as to avoid scrolling when not necessary.

plugins/fontpreview.lua
thacuber2a03: you might want to open a font that's outside the project folder (for whatever reason), so it shows the full path, and a relative path if the font is otherwise in the project

plugins/fontpreview.lua
thacuber2a03: remind me why does `SCALE` exist in the first place lol

Guldoman: Because your monitor could be scaled.
Or you just want the UI scaled up or down.

thacuber2a03:
I realized I just don't want to scroll the view vertically, lol

plugins/fontpreview.lua
Guldoman: I don't think we have a proper list, and it depends on how `freetype` was compiled, sooooo @takase1121?

thacuber2a03: then it's gotta be most font types lmao

plugins/fontpreview.lua
```diff
@@ -12,18 +12,14 @@ function FontView:new(path)
 	FontView.super.new(self)
 	self.path = path
 	self.fonts = {}
-	for i=1, 8 do self.fonts[i] = renderer.font.load(path, 12+i*7) end
+	for i=1, 8 do self.fonts[i] = renderer.font.load(path, (12+i*7)*SCALE) end
```
thacuber2a03, line 15: whoops, forgot to mention I also multiplied by the scale here

thacuber2a03:
before the merge, I want to add something

plugins/fontpreview.lua
```diff
@@ -0,0 +1,63 @@
...
+			self.position.x - self.scroll.x + style.padding.x, y - self.scroll.y,
+			0, style.padding.y
+		)
+		y = y + font:get_height() / 2
+	end
+
+	self:draw_scrollbar()
+end
+
+local supported_types = { ".ttf", ".otf" }
```
Guldoman, line 45:
```suggestion
local supported_types = { "%.ttf$", "%.otf$" }
```
otherwise it matches anything that contains `#ttf`, or `#oft` where `#` is *any* character.

thacuber2a03: oh, right, thanks, I forgot find uses pattern matching

PR #399, opened 2024-03-27
Update lsp_python to 1.1.356

PerilousBooklet:


Guldoman:
This is not needed anymore, right?

PerilousBooklet:
It's not needed.

PR #414, opened 2024-04-06
Added support for Clojure programming language

RohanVashisht1234:


plugins/language_clojure.lua
```diff
@@ -0,0 +1,53 @@
...
+  files = {
+    "%.clojure$",
+  },
+  patterns = {
+    { pattern = ';;.*',                                   type = 'comment' },
+    { pattern = ';.*',                                    type = 'comment' },
+    { pattern = { '#"', '"', '\\' },                      type = 'string' },
+    { pattern = { '"', '"', '\\' },                       type = 'string' },
+    { pattern = { '"""', '"""', '\\' },                   type = 'string' },
+    { pattern = 'Retention%s+()[%a_][%w_/]*',             type = { 'keyword', 'literal' } },
```
Guldoman, line 16:
```suggestion
    { pattern = 'Retention()%s+()[%a_][%w_/]*',             type = { 'keyword', 'normal', 'literal' } },
```

plugins/language_clojure.lua
```diff
@@ -0,0 +1,53 @@
...
+    { pattern = ':[%a_][%w_/%-]*',                        type = { 'keyword2', 'literal' } },
```
Guldoman, line 17: There is no group here, but `type` is a table.

plugins/language_clojure.lua
```diff
@@ -0,0 +1,53 @@
...
+    { pattern = '[%a_][%w_]*()%.()[%a_][%w_/%-]*',        type = { 'keyword','operator', 'keyword2' } },
+    { pattern = "%(()def()%s+()[%a_][%w_%-]*",            type = { "normal", "keyword", "literal", 'literal' } }, -- tested ok
+    { pattern = "%(()def[%a_][%w_]*()%s+()[%a_][%w_%-]*", type = { "normal", "keyword", "literal", 'literal' } }, -- tested ok
```
Guldoman, line 20: Is this about things that start with `def`? So even things like `def_whatever`?

manifest.json
```diff
@@ -735,6 +736,16 @@
       ],
       "version": "0.2"
     },
+    {
+      "description": "Syntax for the [Clojure programming language](https://clojure.org/).",
```
Guldoman, line 740:
```suggestion
      "description": "Syntax for the [Clojure](https://clojure.org/) programming language",
```

plugins/language_clojure.lua
RohanVashisht1234: yes, anything that starts with def is to be highlighted

plugins/language_clojure.lua
RohanVashisht1234: fixed

Guldoman:
Seems like clojure files mostly use a different file extension than what's being checked here.
Would you mind making another PR adding those?
For example in the LSP repository we use these: https://github.com/lite-xl/lite-xl-lsp/blob/59602bc153b567026acc0822fd535a6d82748ce9/config.lua#L132

RohanVashisht1234:
Yes, I have fixed that

PR #423, opened 2024-04-13
Add F# syntax highlighting support.

PerilousBooklet:


plugins/language_fsharp.lua
```diff
@@ -0,0 +1,108 @@
+-- mod-version:3
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "F#",
+  files = { "%.fsi?$" },
+  comment = "//",
```
Guldoman, line 7: Seems like it supports block comments with `(*` and `*)`.

plugins/language_fsharp.lua
```diff
@@ -0,0 +1,108 @@
...
+    { regex   = "\\:\\s?\\w+",                         type = "keyword2"  }, -- Type
+    { regex   = "\\_(?=\\s?\\:)",                      type = "normal"    }, -- _ should be normal when used as ?
+    { pattern = "[%+%-=/%*%^%%<>!~|&_:]",              type = "operator"  }, -- Operators
+    { regex   = [[\.{2}\<?\s?(?=[\\-]?[a-z0-9])]],     type = "operator"  }, -- Range operators
+    { regex   = [[\-\>(?=\s)]],                        type = "function"  }, -- Function arrow
+    { regex   = "\\w+(?=\\s?[(])",                     type = "function"  }, -- Function without generic type
+    { regex   = "\\w+()\\s?\\[?\\<\\'?\\w+\\>\\]?",    type = { "function", "keyword2" } }, -- Function with generic type
+    { regex   = "\\#\\w+",                             type = "keyword2"  }, -- Load
+    { regex   = "\\'\\w+",                             type = "keyword"   }, -- Special variable
+  },
```
Guldoman, line 26: Should this have the basic catch-all-words pattern?

Guldoman (commented):
Needs a version bump for `meta_languages`.

plugins/language_fsharp.lua
PerilousBooklet: Yes, actually. That pattern is for variables that begin with `'`.

plugins/language_fsharp.lua
PerilousBooklet: Thanks, I'll add it right away.

plugins/language_fsharp.lua
Guldoman: I meant that this syntax should have a catch-all at the end, not that regex specifically.

PerilousBooklet: I don't think it need one.

Guldoman: What is matching with the symbols in the symbols table?

PerilousBooklet: Ah wait, it needs one. Because the symbols are words.

PerilousBooklet: Thanks.

Guldoman: No worries.

plugins/language_fsharp.lua
```diff
@@ -0,0 +1,145 @@
+-- mod-version:3
+local syntax = require "core.syntax"
+
+-- Language Syntax References
+-- https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/
+
+syntax.add {
+  name = "F#",
+  files = { "%.fsi?$" },
```
moooodie, line 9: this doesn't handle `.fsx` files

PR #430, opened 2024-04-22
Add PLC Structured Text IEC 61131-3 standard syntax highlighting support.

PerilousBooklet:
You can test this PR with:
- `git clone https://github.com/PerilousBooklet/lite-xl-plugins.git`
- `git fetch && git checkout PR_language_st`
- `lpm run --ephemeral ./ language_st`

You can find the example file used to check the syntax [here](https://drive.google.com/drive/folders/1gyWW2rxrPlzmPfG_-plUemNG0GavRa7-?usp=sharing)


PerilousBooklet:
This PR should be ready for review, but it lacks a couple patterns that require look-behind regex support.

PerilousBooklet:
Fixing REDO vulnerable regex patterns.

plugins/language_st.lua
```diff
@@ -0,0 +1,127 @@
...
+  name = "PLC Structured Text IEC 61131-3",
+  files = { "%.stx?$", "%.iecst$" },
+  comment = "//",
+  block_comment = { "%(%*", "%*%)" },
+  patterns = {
+    { pattern = "//.*",                                     type = "comment"  }, -- Single-line comment
+    { pattern = { "%(%*", "%*%)" },                         type = "comment"  }, -- Multi-line comment
+    { pattern = { '"', '"', '\\' },                         type = "string"   }, -- String, quotation marks
+    { pattern = { "'", "'", '\\' },                         type = "string"   }, -- String, apices
+    { regex   = "\\s\\w+\\#[0-9]+[m|s]?",                   type = "number"   }, -- Time/Date formats
```
Guldoman, line 17:
```suggestion
    { regex   = [[\w+#[0-9]+[ms]?]],                        type = "number"   }, -- Time/Date formats
```

plugins/language_st.lua
```diff
@@ -0,0 +1,127 @@
...
+    { regex   = "(?>[^(\\W]+)(?=\\()",                      type = "function" }, -- Function
```
Guldoman, line 18:
```suggestion
    { regex   = [[(?>[^(\W]+)(?=\()]],                      type = "function" }, -- Function
```

plugins/language_st.lua
```diff
@@ -0,0 +1,127 @@
...
+    { pattern = "[%a_][%w_]*",                              type = "symbol"   }, -- Symbols
+    { regex   = "^\\s*[A-Z]+_[A-Z]*\\s?\\;?\\s?",           type = "keyword"  }, -- Keyword
```
Guldoman, line 20:
```suggestion
    { regex   = [[^\s*[A-Z]+_[A-Z]*\s*;?]],                 type = "keyword"  }, -- Keyword
```
Is this intended to be available only at the start of the line?
Is the underscore in the first pattern supposed to be there, outside the character class?

plugins/language_st.lua
```diff
@@ -0,0 +1,127 @@
...
+    { regex   = "\\:\\s*\\w+",                              type = "keyword2" }, -- Variable/Method Type
+    { regex   = "(?:\\-|\\+)?\\d+\\.*(?:\\-|\\+)?\\d+",     type = "number"   }, -- Number Range
```
Guldoman, line 22:
```suggestion
    { regex   = [[:\s*\w+]],                                type = "keyword2" }, -- Variable/Method Type
    { regex   = [[[-+]?\d+\.*[-+]?\d+]],                    type = "number"   }, -- Number Range
```
In the second regex, is it intended to allow matching any number of `.`, or should it only match `..`?

plugins/language_st.lua
```diff
@@ -0,0 +1,127 @@
...
+    { pattern = "[%+%-=/%*%^%%<>!~|&:]",                    type = "operator" }, -- Operators
+    { pattern = "-?0x%x+",                                  type = "number"   }, -- Number
+    { pattern = "-?%d+[%deE]*f?",                           type = "number"   }, -- Number
+    { pattern = "-?%.?%d+f?",                               type = "number"   }, -- Number
+    { regex   = "^\\w+()\\s(?:[A-Z]+[a-z]*\\_?)+(?=\\w*)",  type = { "normal", "function" } }, -- Function implementation
```
Guldoman, line 27:
```suggestion
    { regex   = [[^\w+()\s(?:[A-Z]+[a-z]*\_?)+(?=\w*)]],    type = { "normal", "function" } }, -- Function implementation
```
Should this allow for spaces at the start?
What about allowing multiple spaces between the first `\w+` and the group?
What is the group trying to match?
What is the lookahead at the end doing?

plugins/language_st.lua
PerilousBooklet: Yes and yes. It needs to match things like `TIME_AND_DATE` and `DATE`, at the beginning of the line.

plugins/language_st.lua
PerilousBooklet: It should also match `...` other than `..`.

plugins/language_st.lua
Guldoman: That regex is not doing that, it's only matching one underscore, and it's only matching when the underscore is present.

PerilousBooklet: Actually I think this regex pattern isn't even needed, because it should do what the symbol lua pattern (`[%a_][%w_]*`) does.
And the keywords are highlighted even without this regex pattern.
I don't remember why it was there honestly, it's been too long.

plugins/language_st.lua
PerilousBooklet: This pattern as well, for some reason removing it doesn't affect any highlighted example of code.

plugins/language_st.lua
```diff
@@ -0,0 +1,125 @@
...
+  files = { "%.stx?$", "%.iecst$" },
+  comment = "//",
+  block_comment = { "%(%*", "%*%)" },
+  patterns = {
+    { pattern = "//.*",                                     type = "comment"  }, -- Single-line comment
+    { pattern = { "%(%*", "%*%)" },                         type = "comment"  }, -- Multi-line comment
+    { pattern = { '"', '"', '\\' },                         type = "string"   }, -- String, quotation marks
+    { pattern = { "'", "'", '\\' },                         type = "string"   }, -- String, apices
+    { pattern = "%w+#[0-9]+m?s?",                           type = "number"   }, -- Time/Date formats
+    { regex   = [[(?>[^(\W]+)(?=\()]],                      type = "function" }, -- Function
```
Guldoman, line 18: Could this be a simple "standard" function pattern?

PerilousBooklet: Yes, probably.

plugins/language_st.lua
```diff
@@ -0,0 +1,125 @@
...
+  block_comment = { "%(%*", "%*%)" },
+  patterns = {
+    { pattern = "//.*",                                     type = "comment"  }, -- Single-line comment
+    { pattern = { "%(%*", "%*%)" },                         type = "comment"  }, -- Multi-line comment
+    { pattern = { '"', '"', '\\' },                         type = "string"   }, -- String, quotation marks
+    { pattern = { "'", "'", '\\' },                         type = "string"   }, -- String, apices
+    { pattern = "%w+#[0-9]+m?s?",                           type = "number"   }, -- Time/Date formats
+    { pattern = "[%a_][%w_]*%f[(]",                         type = "function" }, -- Function
+    { pattern = "[%a_][%w_]*",                              type = "symbol"   }, -- Symbols
+    { regex   = "\\:\\s*(?:\\w+_?)+",                       type = "keyword2" }, -- Variable/Method Type
```
Guldoman, line 20: Would
```suggestion
    { pattern = ":%s*%w[%w_]*",                             type = "keyword2" }, -- Variable/Method Type
```
be equivalent?

PerilousBooklet: Yes, thanks!

manifest.json
```diff
@@ -168,7 +169,7 @@
       "id": "meta_languages",
       "mod_version": "3",
       "type": "meta",
-      "version": "0.1.18"
+      "version": "0.1.19"
```
Guldoman, line 172:
```suggestion
      "version": "0.1.20"
```

PR #432, opened 2024-05-01
Add Vala syntax highlighting.

PerilousBooklet:
You can test this PR with:
- `git clone https://github.com/PerilousBooklet/lite-xl-plugins.git`
- `git fetch && git checkout PR_language_vala`
- `lpm run --ephemeral ./ language_vala`

You can find the example file used to check the syntax [here](https://drive.google.com/drive/folders/1gyWW2rxrPlzmPfG_-plUemNG0GavRa7-?usp=sharing)

### DIFF
--- a/plugins/language_vala.lua
+++ b/plugins/language_vala.lua
@@ -4,9 +4,6 @@ local syntax = require "core.syntax"
 -- Language syntax reference
 -- https://gnome.pages.gitlab.gnome.org/vala/manual/overview.html
 
--- https://devina.io/redos-checker
--- https://redosdetector.com/?pattern=%5C%40%3F%5C%22&caseInsensitive=false&unicode=false
-
 syntax.add {
   name = "Vala",
   files = { "%.vala$" },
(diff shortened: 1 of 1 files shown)


plugins/language_vala.lua
```diff
@@ -0,0 +1,121 @@
...
+    { regex   = "\\w+(?=\\s+\\w++\\s+\\=\\s+)",                                   type = "function" }, -- Class name in class instance
+    { regex   = "\\<.+\\>",                                                       type = "keyword2" }, -- Generic Type
+    { pattern = "[%+%-=/%*%^%%<>!~|&]",                                           type = "operator" }, -- ?
+    { pattern = "[%a_][%w_]*%f[(]",                                               type = "function" }, -- ?
+    { pattern = "[%a_][%w_]*",                                                    type = "symbol"   }, -- ?
+    { regex   = "\\s?\\:\\s?(?:\\w+\\.?)+(?=\\s?[{])",                            type = "keyword2" }, -- Inheritance
+    { regex   = "\\s?\\:\\s?(?:\\w+(?:\\<.+\\>)?\\s?\\,?\\s?)+(?=\\s?[{])",       type = "keyword2" }, -- Inheritance
+    { regex   = "class()\\s\\w+()\\<.+\\>(?=\\s?\\:)",                            type = { "keyword", "normal", "keyword2" } }, -- Generic Class Type
+    { regex   = "interface()\\s\\w+()\\<.+\\>(?=\\s?\\:)",                        type = { "keyword", "normal", "keyword2" } }, -- Generic Interface Type
```
Guldoman, line 25: Should the space matches allow for multiple spaces?

Also you might want to use raw strings to avoid all the double escaping.

plugins/language_vala.lua
```diff
@@ -0,0 +1,121 @@
...
+    { regex   = "\\=\\>(?=[{])",                                                  type = "keyword"  }, -- Lambda
+    { regex   = "[A-Z][A-Z_]+",                                                   type = "keyword2" }, -- Constants
+    { regex   = "\\[.+\\](?=\\w*)",                                               type = "keyword2" }, -- Attribute
+    { regex   = "\\#[if|elif|else|endif](?=\\s?\\w*)",                            type = "keyword"  }, -- Preprocessor directive
```
Guldoman, line 29: I don't think this regex is correct.
`[if|elif|else|endif]` will not match the words `if`, `elif`, `else` and `endif`.

plugins/language_vala.lua
```diff
@@ -0,0 +1,121 @@
...
+    { pattern = { '"', '"', '\\' },                                               type = "string"   }, -- ?
+    { pattern = { "'", "'", '\\' },                                               type = "string"   }, -- ?
+    { pattern = "-?0x%x+",                                                        type = "number"   }, -- ?
+    { pattern = "-?%d+[%d%.eE]*f?",                                               type = "number"   }, -- ?
+    { pattern = "-?%.?%d+f?",                                                     type = "number"   }, -- ?
+    { regex   = "\\w+(?=\\s+\\w++\\s+\\=\\s+)",                                   type = "function" }, -- Class name in class instance
+    { regex   = "\\<.+\\>",                                                       type = "keyword2" }, -- Generic Type
+    { pattern = "[%+%-=/%*%^%%<>!~|&]",                                           type = "operator" }, -- ?
+    { pattern = "[%a_][%w_]*%f[(]",                                               type = "function" }, -- ?
+    { pattern = "[%a_][%w_]*",                                                    type = "symbol"   }, -- ?
```
Guldoman, line 21: This is the same as the pattern at the end.

plugins/language_vala.lua
PerilousBooklet: The space matches could be more than one I guess, but they shouldn't be there.

plugins/language_vala.lua
PerilousBooklet: I changed the pattern, thanks.

PR #434, opened 2024-05-03
Added Conf language

radupotop:
Syntax for `.gitconfig`, `.rclone.conf` and other generic configuration files based on the INI format.

### DIFF
--- a/plugins/language_conf.lua
+++ b/plugins/language_conf.lua
@@ -1,0 +1,38 @@
+-- mod-version:3
+
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "Conf",
+  files = { ".rclone.conf", ".gitconfig", "%.conf$" },
+  comment = '#',
+  patterns = {
+    { pattern = ";.-\n", type = "comment" },
+    { pattern = "#.-\n", type = "comment" },
+    { pattern = { "%[", "%]" }, type = "keyword" },
+
+    { pattern = { '"""', '"""', '\\' }, type = "string" },
+    { pattern = { '"', '"', '\\' }, type = "string" },
+    { pattern = { "'''", "'''" }, type = "string" },
+    { pattern = { "'", "'" }, type = "string" },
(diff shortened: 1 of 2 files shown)


plugins/language_conf.lua
```diff
@@ -0,0 +1,46 @@
...
+
+    { pattern = { '"""', '"""', '\\' }, type = "string" },
+    { pattern = { '"', '"', '\\' }, type = "string" },
+    { pattern = { "'''", "'''" }, type = "string" },
+    { pattern = { "'", "'" }, type = "string" },
+    { pattern = "[A-Za-z0-9_%.%-]+%s*%f[=]", type = "function" },
+    { pattern = "%s+%-%-[A-Za-z0-9%-]+", type = "normal" },
+    -- Git
+    { pattern = "HEAD", type = "literal" },
+    -- Rclone AWS S3 canned ACLs
```
Guldoman, line 22: I'm wondering if there should be multiple syntaxes at this point.

radupotop: Thanks for the review.  
They could be separate syntax files, but at the same time all these conf files are very alike – just INI files with slightly different keywords.

plugins/language_conf.lua
```diff
@@ -0,0 +1,46 @@
+-- mod-version:3
+
+local syntax = require "core.syntax"
+
+syntax.add {
+  name = "Conf",
+  files = { ".gitconfig", ".rclone.conf", "%.conf$" },
```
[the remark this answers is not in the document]
radupotop: Actually in the case of `.gitconfig` you sometimes want to have multiple `.gitconfig*` files so that you can use different 'personas' depending on the project, e.g.:

`.gitconfig`
```
[user]
    email = mrhackerman@aol.com
    signingkey = 123
[includeIf "gitdir:~/WorkDir/"]
    path = .gitconfig-work
```

`.gitconfig-work`
```
[user]
    email = johndoe@ibm.com
    signingkey = 456
```

PR #439, opened 2024-05-25
add sectioning commands for ConTeXt

ousia:


PerilousBooklet:
You can use a `non-capturing group` to avoid repeating the subsub... strings, like this 
image

PerilousBooklet:
Yes I didn't notice, you can do `\\(?:sub)*section`, but for the paragraph `\\(?:sub)?paragraph` is better (if there are only paragraph and subparagraph).

PerilousBooklet:
Wait why a new PR?

PerilousBooklet:
I use this to test regex: https://regex101.com/

PerilousBooklet:
By the way, if you use regex you must use `regex` instead of `pattern`.
`pattern` is for lua patterns, not regex patterns.

PR #441, opened 2024-05-31
eval:selected multi-cursor support

exelotl:
This PR makes it so that if you run `eval:selected` while having multiple selections, it will evaluate and replace the contents of each, rather than replacing all selections with the first result.

Note: some previous functionality:

> if no text is selected, eval the current line and insert the result in the next line with "= " prefix. (https://github.com/lite-xl/lite-xl-plugins/pull/177)

has been removed as it doesn't play nice with multiple selections. Now it behaves as if the whole line had been selected, and replaces the line. Maybe @Titousensei could confirm that this is ok / won't ruin your workflow?

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -422,7 +422,7 @@
       "id": "eval",
       "mod_version": "3",
       "path": "plugins/eval.lua",
-      "version": "0.1"
+      "version": "0.2"
     },
     {
       "description": "Adds Treesitter syntax highlighting support",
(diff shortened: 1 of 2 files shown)


adamharrison:
Sorry about the delay.

This works for me! Looks good. Can you bump the version in the manifest, please? I'll merge after that.

adamharrison:
Actually, I can do this for you. Done. Merging.

PR #442, opened 2024-06-05
updates gitblame remote URL to point to the commit with the correct version

juliardi:
When running LPM to install a package, it output a warning "warning: gitblame stub on https://github.com/lite-xl/lite-xl-plugins.git:master has differing version from remote (0.2 vs 0.1); may lead to install being inconsistent". This PR should fix that.


juliardi:
Alright, I have bumped it to 0.2.1

Guldoman:
Looks good! Thank you!

PR #446, opened 2024-06-24
Fix codeplus mod_version

DazFather:
As explicitly written on the specification mod_version is a string and  I think that is better to not introduce more ambiguity on the manifest in order to simplify parsing, even if I personally prefer mod_version as a simple incremental integer number instead of a string, I think it would be better to encourage a stronger standard


Guldoman:
Yeah, that's fair. Thank you!

PR #448, opened 2024-07-02
Add Nim string interpolation syntax highlighting

carrexxii:
Screenshot_2024-07-02_23-05-04

### DIFF
--- a/plugins/language_nim.lua
+++ b/plugins/language_nim.lua
@@ -101,7 +101,7 @@ end
 
 local interpolation_syntax = {
   { pattern = { '{{', '}}' }, type = "string" },
-  { pattern = { '{', '}', '\\' }, type = "keyword2", syntax = { patterns = patterns, symbols = symbols, } },
+  { pattern = { '{', '}', '\\' }, type = "keyword2", syntax = ".nim" },
   { pattern = "[%S][%w]*", type = "string" },
 }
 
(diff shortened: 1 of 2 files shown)


plugins/language_nim.lua
```diff
@@ -88,14 +88,37 @@ for _, type in ipairs(standard_generic_types) do
...
+    syntax  = syntax and {
+      patterns = syntax,
+      symbols  = {},
+    } or nil
+  }
+end
+
+local interpolation_syntax = {
+  { pattern = { '{{', '}}' }, type = "string" },
+  { pattern = { '{', '}', '\\' }, type = "keyword2", syntax = { patterns = patterns, symbols = symbols, } },
```
Guldoman, line 104: Would this be equivalent to:
```suggestion
  { pattern = { '{', '}', '\\' }, type = "keyword2", syntax = ".nim" },
```

plugins/language_nim.lua
```diff
@@ -88,14 +88,37 @@ for _, type in ipairs(standard_generic_types) do
...
   -- strings and chars
-  { pattern = { '"', '"', '\\' },             type = "string" },
-  { pattern = { '"""', '"""[^"]' },           type = "string" },
   { pattern = { "'", "'", '\\' },             type = "literal" },
+  string_pattern('"""', '"""[^"]'),
+  string_pattern('"'  , '"'      ),
+  string_pattern('\\"', '\\"'    ), -- For highlighting strings inside iterpolated blocks
+  -- string interpolation
+  string_pattern('%&"""' , '"""[^"]', interpolation_syntax),
+  string_pattern('fmt"""', '"""[^"]', interpolation_syntax),
```
Guldoman, line 119: `[^"]` is a bit problematic, as it captures the character.
Look into either using Lua frontier patterns or regex with negative lookaheads.

carrexxii: Ok, I'll have a look. I just kept it from the original syntax

carrexxii: I made the change for frontier matches as I understand them and have the same result as in my original image

Guldoman:
Looks good!
Could you also bump the plugin version in `manifest.json`?

carrexxii:
I bumped the version

Guldoman:
Looks good, thank you!

PR #450, opened 2024-07-06
GitLab Markdown Preview Plugin

ghost:
Hello, I would like to contribute the following plugin. It renders Markdown through the GitLab API, similarly got ghmarkdown.lua.
I hope this PR follows your requirements.
I am happy to take any kind of improvements and suggestions.


Guldoman:
I wonder if the plugins should be merged, considering how similar they are. What do you think?

PerilousBooklet:
@notvanessa You could add a configuration option to `ghmarkdown` to switch from Github's API to Gitlab's API.

notvanessa:
I would like to keep it separated it is just cleaner this way, but I don't use the plugin anymore so feel free to close the PR

PR #451, opened 2024-07-10
Update nerdicons

Gaspartcho:
This small PR adds some more icons to the `nerdicons` plugin. 

Those icons are:

- julia files
- mp4 files
- jld2 (subset of the HDF5 file format for julia)
- jupyter notebook files (I did not find the jupyter logo in the nerdfonts library so instead I put the symbol of a regular notebook)


Guldoman:
Looks good, thank you!

PR #453, opened 2024-07-10
sync

Mandarancio:


plugins/language_marte.lua
```diff
@@ -0,0 +1,42 @@
...
+    },
+    {
+      pattern = "Type%s+()=()%s+[%a_][%w_]*",
+      type = { "keyword", "operator", "keyword2"}
+    },
+    {
+      pattern = "[%+%$][%a_][%w_]+%s()=",
+      type = {"function", "operator"}
+    },
+    { pattern = "=%s+()[%a_][%w_]+", type = "string" },
```
Guldoman, line 28: This is dividing the match in two `()`, but only a single token type is provided.

README.md
Guldoman, line 1: No need to update the README, it'll be automatically updated from the manifest info.

Guldoman (commented):
Could you also add the new languages as dependencies of `meta_languages`?
Bump its version too.

Guldoman:
Wait, I'm a bit confused about what's going on here...

Still, the comment about the MARTe language stands.

Mandarancio:
Sorry my bad, I did not want to publish this pull request, I will delete it...

PR #454, opened 2024-07-10
added keyhud plugin

Mandarancio:
added new plugin: [keyhud](https://codeberg.org/Mandarancio/keyhud)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -608,7 +608,7 @@
       "id": "keyhud",
       "description": "Simple *key HUD* for Lite XL",
       "mod_version": "3",
-      "remote": "https://codeberg.org/Mandarancio/keyhud:984f61c56e409ad62b92dff6fa23f399f761d9ab",
+      "remote": "https://codeberg.org/Mandarancio/keyhud:1560d11fb5485aea264ea742e7ee6cd7693f03e0",
       "type": "plugin",
       "version": "0.1"
     },
--- a/manifest.json
+++ b/manifest.json
@@ -606,7 +606,7 @@
     },
     {
       "id": "keyhud",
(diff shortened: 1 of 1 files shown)


manifest.json
```diff
@@ -604,6 +604,14 @@
       "type": "library",
       "version": "22.0.1"
     },
+    {
+      "id": "keyhud",
+      "description": "Simple *key HUD* for lite-xl"
```
Guldoman, line 609: Missing comma.
Same thing in the manifest in your repo.

Also, might be better to use `Lite XL` instead of `lite-xl`.

Mandarancio: Thank you very much, I did not pay attention when I created the manifest on my repo and then copy pasted it... sorry

I changed `lite-xl` to `Lite XL` :)

manifest.json
```diff
@@ -604,6 +604,14 @@
       "type": "library",
       "version": "22.0.1"
     },
+    {
+      "id": "keyhud",
+      "description": "Simple *key HUD* for Lite XL",
+      "mod_version": "3",
+      "remote": "https://codeberg.org/Mandarancio/keyhud:984f61c56e409ad62b92dff6fa23f399f761d9ab",
```
Guldoman, line 611: You also need to update the commit hash, or `lpm` will try to use the old version with the missing comma:
```suggestion
      "remote": "https://codeberg.org/Mandarancio/keyhud:1560d11fb5485aea264ea742e7ee6cd7693f03e0",
```

Guldoman:
Looks awesome!

Only suggestion is that for configurable things, you should use `config.plugins.keyhud` with `common.merge` like how most plugins do it.

PR #456, opened 2024-07-12
Update keyhud version

Mandarancio:
Update keyhud version to use `config.plugins.keyhud` table for configuration

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -608,9 +608,9 @@
       "description": "Simple *key HUD* for Lite XL",
       "id": "keyhud",
       "mod_version": "3",
-      "remote": "https://codeberg.org/Mandarancio/keyhud:1560d11fb5485aea264ea742e7ee6cd7693f03e0",
+      "remote": "https://codeberg.org/Mandarancio/keyhud:a029068578bc582ea85f8204199818be26a3d397",
       "type": "plugin",
-      "version": "0.1"
+      "version": "0.2"
     },
     {
       "description": "Exports the keymap to a JSON file.",


Guldoman:
Looks good.

I noticed a little regression, tho. A few icons are broken now, like caps lock, backspace, delete... :cry:.
Do you want to merge this anyways for now?

Guldoman:
Oh I also noticed, you might not be using `common.merge` correctly.
It should be used like this:
```lua
common.merge({
  -- your default config
}, config.plugins.keyhud)
```

Mandarancio:
> Looks good.
> 
> I noticed a little regression, tho. A few icons are broken now, like caps lock, backspace, delete... 😢. Do you want to merge this anyways for now?

I switched to the  default `code_font` to have monospaced icon but the default font does not have all the glyphs... 
It is possible to configure any font in the user module (`init.lua`).

I am working in the new version with more features (possibility to scale size of the hud, positioning it at the left side, etc)

Guldoman:
Thank you!

Feature request: show a "multiplier" near the key name when it's pressed multiple times; or even an option to allow duplicate key names to appear in the hud.

Mandarancio:
Very good idea about the multiplayer, as well  I was thinking to keep the
key on display till it is pressed (and let the timer start after it is
released) what do you think?

Guldoman:
> I was thinking to keep the key on display till it is pressed (and let the timer start after it is released) what do you think?

Yeah makes sense. Should it also be highlighted until it's released?
Or even show an arrow down on press, and up on release.

PR #457, opened 2024-07-16
Update language julia

Gaspartcho:
Fixed some bugs / added some improvements here and there.

Added syntax highlighting for string interpolation (see picture bellow)

image


plugins/language_julia.lua
```diff
@@ -4,109 +4,130 @@
...
+    ["Float16"]      = "keyword2",
+    ["Float32"]      = "keyword2",
+    ["Float64"]      = "keyword2",
+    ["Vector"]       = "keyword2",
+    ["Matrix"]       = "keyword2",
+    ["Ref"]          = "keyword2",
+    ["String"]       = "keyword2",
+    ["Function"]     = "keyword2",
+    ["Number"]       = "keyword2",
+    ["im"]           = "keyword2",
```
Guldoman, line 123: Is this a keyword?

Gaspartcho: I'm not sure, it was like that before, so I just left it there. 

I can change it though

PR #458, opened 2024-07-17
update language_toml

Gaspartcho:
Mainly formatted the file, simplified the number declaration (most of `[0-9]` could be replaced by `%d`) as well as the tables part.

Also added a minimal date highlighting (by highlighting some special characters used in dates).

Source for the documentation: https://toml.io/en/v1.0.0

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1600,7 +1600,7 @@
       "tags": [
         "language"
       ],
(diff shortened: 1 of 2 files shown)


Guldoman:
Looks good, thank you!

PR #459, opened 2024-07-18
Add plugin `previewer`

Mandarancio:
Simple plugin that allow to preview file and buffers before before opening them.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1959,6 +1959,14 @@
       "version": "0.1"
     },
     {
+      "id": "previewer",
+      "description": "Preview files and doc before before opening them",
+      "mod_version": "3",
+      "remote": "https://codeberg.org/Mandarancio/lite-previewer:49a4747279276bb40e993fabb726c0cb43fa0899",
+      "type": "plugin",
+      "version": "0.1"
+    },
+    {
       "description": "Adds middle mouse click copy/paste (primary selection). To use this plugin, `xclip` must be installed.",
       "id": "primary_selection",
       "mod_version": "3",
--- a/manifest.json
+++ b/manifest.json
@@ -1962,7 +1962,7 @@
       "id": "previewer",
       "description": "Preview files and doc before before opening them",
       "mod_version": "3",
-      "remote": "https://codeberg.org/Mandarancio/lite-previewer:659ce9b8c33d57841bebc956b373e2726dd86586",
+      "remote": "https://codeberg.org/Mandarancio/lite-previewer:39c6c4667009f669bdc84a99360455c8e384bd4d",
       "type": "plugin",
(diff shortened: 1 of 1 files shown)


Guldoman:
Mmmh as it is, this plugin is a bit problematic as it overrides the CommandView in every case, so it'll try previewing anything the CommandView is showing.
It even causes a crash in some cases, for example when trying to create a new file from the context menu of the TreeView (only if you try doing that with one of the first directories for some reason).

A better solution would be to provide your own function to open files. I think it could use `suggest` directly to change files.
You could also provide a wrapper function that will have to be passed to `suggest`, for people/plugins that want to integrate your preview.

As a note, you don't really need to specify `.` as path in the manifest.

Mandarancio:
Thank you for your feedback! It is really nice to have someone testing it,
as I never encountered issues but is because I never use the tree view and
never create a new file from lite . I will fix it as you suggested!

Mandarancio:
> Mmmh as it is, this plugin is a bit problematic as it overrides the CommandView in every case, so it'll try previewing anything the CommandView is showing. It even causes a crash in some cases, for example when trying to create a new file from the context menu of the TreeView (only if you try doing that with one of the first directories for some reason).

I fixed the issue by creating a new method called `enter_with_preview` so that you must explicitly call it for enabling the preview.
Also I overwritten `core:open-file` and `core:find-file` to use it.

> A better solution would be to provide your own function to open files. I think it could use `suggest` directly to change files. You could also provide a wrapper function that will have to be passed to `suggest`, for people/plugins that want to integrate your preview.

The problem that `suggest` will be only called when some input text is changed  and not if you move your arrows around.
To get the correct `suggestion_idx` the only way I found was to overwrite `CommandView:update`. 

> As a note, you don't really need to specify `.` as path in the manifest.

Done

adamharrison:
> The problem that suggest will be only called when some input text is changed and not if you move your arrows around.
To get the correct suggestion_idx the only way I found was to overwrite CommandView:update.

I think this is fine for now, but it makes me think we should have an additional callback in `CommandView:enter` to trigger on suggestion selection, like `change`.

> I fixed the issue by creating a new method called enter_with_preview so that you must explicitly call it for enabling the preview. Also I overwritten core:open-file and core:find-file to use it.

I think this is also reasonable. I like it.

adamharrison (approved):

adamharrison:
Of note, this will crash on certain binary files on `master`, but that's not necessarily a problem with this plugin, more with lite itself.

Discussion on discord was that this is probably OK; rather than handling it in this plugin, there are numerous ways to treat opening a binary file, including plugins that throw up confirmation screens, so I think this is fine. Merging.

PR #460, opened 2024-07-18
Add plugin `buffers`

Mandarancio:
Add `buffers` plugins that enable to manage documents (buffers) and not tabs in Lite XL


Mandarancio:
Converting to draft as I will update it to match new `previewer` interface

Mandarancio (commented):
Ok for me

Issue #469, opened 2024-07-29
Some issues with `language_sh`


Samueru-sama:
* Scripts that start with a `#!/bin/sh -e` don't get recognized and have no highlighting as result.

For example, here is part of the `AppRun` script of the suyu appimage:

![image](https://github.com/user-attachments/assets/06dfa69d-c68d-4fdf-866e-7e1035ddea04)

Removing the `-e` flag from the sheabang fixes the issue. 

* I haven't figured out why this happens, but the [functions in the bemoji script are all highlighted with the same color](https://github.com/marty-oehme/bemoji/blob/main/bemoji)

![image](https://github.com/user-attachments/assets/242e0456-6145-42e9-84ec-0d8fe1cbc0b8)

* And finally files like `.bash_profile`, `.zprofile`, etc never get hightlited as they normally don't have a sheabang. 

[What's odd is that there is actually support for these files in the script so I don't know why it doesn't work](https://github.com/lite-xl/lite-xl-plugins/blob/master/plugins/language_sh.lua#L6)

Samueru-sama:
I've discovered that it works if the file is named: 

`.sh`
`.bash`
`.zsh`
`.zshrc`
`.fish`

however it doesn't work when it is named: 

`.bashrc`
`.bash_profile`
`.profile`
`.zprofile` (Even after adding it to the files list).

And `.zshrc` also doesn't work, however that is because it wasn't added to the list of files, unlike `.zprofile` it does work after being added.

PR #470, opened 2024-07-29
Add .zprofile and .zshrc to `language_sh` and fix header pattern.

Samueru-sama:
Now `.zshrc` gets recognized and scripts that have flags in their shebang also work.

However my `.zprofile` still doesn't work 🤔

When completed this will fix all the issues at #469

### DIFF
--- a/plugins/language_sh.lua
+++ b/plugins/language_sh.lua
@@ -3,8 +3,8 @@ local syntax = require "core.syntax"
 
 syntax.add {
   name = "Shell script",
-  files = { "%.sh$", "%.bash$", "^%.bashrc$", "^%.bash_profile$", "^%.profile$", "%.zsh$", "%.fish$" },
-  headers = "^#!.*bin.*sh\n",
+  files = { "%.sh$", "%.bash$", "^%.bashrc$", "^%.bash_profile$", "^%.profile$", "^%.zprofile$", "%.zsh$", "%.zshrc$", "%.fish$" },
+  headers = "^#!.*bin.*sh",
   comment = "#",
   patterns = {
     -- $# is a bash special variable and the '#' shouldn't be interpreted


Samueru-sama:
Alright I think this is ready now. 

I'm not very sure how [this](https://github.com/lite-xl/lite-xl-plugins/pull/470/commits/26a5567134effa790b7c39c30bec243538d9a622) fixed this issue I was having [here](https://github.com/lite-xl/lite-xl-plugins/issues/469#issuecomment-2256426110), but it works lol

I've also tested it with other scripts to make sure this doesn't introduce a bug and haven't noticed anything odd.

Guldoman:
> I'm not very sure how [this](https://github.com/lite-xl/lite-xl-plugins/pull/470/commits/26a5567134effa790b7c39c30bec243538d9a622) fixed this issue I was having [here](https://github.com/lite-xl/lite-xl-plugins/issues/469#issuecomment-2256426110), but it works lol

It's because of nested `${...}`.
With `.-` it takes as few characters as possible, with `.*` it takes as many as possible.
So with `.-`, it stops at the first `}`, while with `.*` it stops at the last.
This is problematic, because it will now match things like `${wq${as}asd} asdasd ${saasasd}` wrongly.

What do you think about something like:
```diff
index 83e98b5..8396464 100644
--- a/plugins/language_sh.lua
+++ b/plugins/language_sh.lua
@@ -1,6 +1,16 @@
 -- mod-version:3
 local syntax = require "core.syntax"
 
+local string_syntax = {
+  name = "Shell script string",
+  patterns = {
+    { pattern = {"%${", "}"}, type = "keyword2", syntax = ".sh" },
+    { pattern = "[^%$]",      type = "string" },
+    { pattern = "%$[^{]",     type = "string" },
+  },
+  symbols = {}
+}
+
 syntax.add {
   name = "Shell script",
   files = { "%.sh$", "%.bash$", "%.bashrc$", "%.bash_profile$", "%.profile$", "%.zprofile$", "%.zsh$", "%.zshrc$", "%.fish$" },
@@ -11,11 +21,11 @@ syntax.add {
     -- as a comment.
     { pattern = "$[%a_@*#][%w_]*",                type = "keyword2" },
     -- Comments
-    { pattern = "#.*\n",                          type = "comment"  },
+    { pattern = "#.*",                            type = "comment"  },
     -- Strings
-    { pattern = { '"', '"', '\\' },               type = "string"   },
-    { pattern = { "'", "'", '\\' },               type = "string"   },
-    { pattern = { '`', '`', '\\' },               type = "string"   },
+    { pattern = { '"', '"', '\\' },               type = "string", syntax = string_syntax },
+    { pattern = { "'", "'", '\\' },               type = "string"                         },
+    { pattern = { '`', '`', '\\' },               type = "string", syntax = ".sh"         },
     -- Ignore numbers that start with dots or slashes
     { pattern = "%f[%w_%.%/]%d[%d%.]*%f[^%w_%.]", type = "number"   },
     -- Operators
@@ -35,12 +45,14 @@ syntax.add {
     -- Match variable assignments
     { pattern = "[_%a][%w_]+%f[%+=]",              type = "keyword2" },
     -- Match variable expansions
-    { pattern = "${.*}",                           type = "keyword2" },
+    { pattern = { "${", "}", "\\" },               type = "keyword2", syntax = ".sh"      },
     { pattern = "$[%d$%a_@*][%w_]*",               type = "keyword2" },
     -- Functions
     { pattern = "[%a_%-][%w_%-]*[%s]*%f[(]",       type = "function" },
     -- Everything else
     { pattern = "[%a_][%w_]*",                     type = "symbol"   },
+    -- Catch escaped } - to avoid closing a ${...\}...} prematurely
+    { pattern = "\\}",                             type = "normal"   },
   },
   symbols = {
     ["case"]      = "keyword",
```

Also, would you mind bumping the plugin version in `manifest.json`?

Samueru-sama:
>It's because of nested ${...}.
With .- it takes as few characters as possible, with .* it takes as many as possible.
So with .-, it stops at the first }, while with .* it stops at the last.
This is problematic, because it will now match things like ${wq${as}asd} asdasd ${saasasd} wrongly.

Hi, thanks for the explanation but I'm still not sure why it fixes the issue 😅

I know that `.*` will match more than `.-`, but if that is the case, why does `.-` cause all functions in this [script ](https://github.com/marty-oehme/bemoji/blob/main/bemoji) to be matched? 

I know the problem is line 5 in the script, something happens that makes it seem like the last } wasn't there I think.

If it was the other way around, that .* had the issue and .- fixes it it would make sense, but anyway I just tested the changes you gave (I applied them to the file of my fork just in case) and something broke.

This is how part of the bemoji script looks with the changes I did: 

image

And this is how it looks with the proposed changes: 

image

>Also, would you mind bumping the plugin version in manifest.json?

Will do!

Guldoman:
The issue is caused by `#` here https://github.com/marty-oehme/bemoji/blob/cc9f809446f75a67e2128749537c51dd9558eec8/bemoji#L81.

So I guess we can't just say that inside `${...}` it nests the entire syntax. We'll have to make one on purpose for it.
Something like:
```diff
index 83e98b5..bbd93f1 100644
--- a/plugins/language_sh.lua
+++ b/plugins/language_sh.lua
@@ -1,6 +1,26 @@
 -- mod-version:3
 local syntax = require "core.syntax"
 
+local expansion_syntax = {
+  name = "Shell script parameter expansion",
+  symbols = {}
+}
+expansion_syntax.patterns = {
+    { pattern = { "%${", "}", "\\" }, type = "keyword2", syntax = expansion_syntax },
+    { pattern = "[^%$}]+",            type = "string" },
+    { pattern = "[%$}]",              type = "string" },
+}
+
+local string_syntax = {
+  name = "Shell script string",
+  patterns = {
+    { pattern = { "%${", "}", "\\" }, type = "keyword2", syntax = expansion_syntax },
+    { pattern = '[^%$"]+',            type = "string" },
+    { pattern = '[%$"]',              type = "string" },
+  },
+  symbols = {}
+}
+
 syntax.add {
   name = "Shell script",
   files = { "%.sh$", "%.bash$", "%.bashrc$", "%.bash_profile$", "%.profile$", "%.zprofile$", "%.zsh$", "%.zshrc$", "%.fish$" },
@@ -11,11 +31,11 @@ syntax.add {
     -- as a comment.
     { pattern = "$[%a_@*#][%w_]*",                type = "keyword2" },
     -- Comments
-    { pattern = "#.*\n",                          type = "comment"  },
+    { pattern = "#.*",                            type = "comment"  },
     -- Strings
-    { pattern = { '"', '"', '\\' },               type = "string"   },
-    { pattern = { "'", "'", '\\' },               type = "string"   },
-    { pattern = { '`', '`', '\\' },               type = "string"   },
+    { pattern = { '"', '"', '\\' },               type = "string", syntax = string_syntax },
+    { pattern = { "'", "'", '\\' },               type = "string"                         },
+    { pattern = { '`', '`', '\\' },               type = "string", syntax = ".sh"         },
     -- Ignore numbers that start with dots or slashes
     { pattern = "%f[%w_%.%/]%d[%d%.]*%f[^%w_%.]", type = "number"   },
     -- Operators
@@ -35,7 +55,7 @@ syntax.add {
     -- Match variable assignments
     { pattern = "[_%a][%w_]+%f[%+=]",              type = "keyword2" },
     -- Match variable expansions
-    { pattern = "${.*}",                           type = "keyword2" },
+    { pattern = { "${", "}", '\\' },               type = "keyword2", syntax = expansion_syntax },
     { pattern = "$[%d$%a_@*][%w_]*",               type = "keyword2" },
     -- Functions
     { pattern = "[%a_%-][%w_%-]*[%s]*%f[(]",       type = "function" },
```

Samueru-sama:
That one works fine, thanks! I will apply the patch.

>In the future we might even improve the expansion syntax, to highlight inside it too.

something I would like to see is variables inside `" "` get highlighted a different color, right now if you have a path with a variable it all has the same color.

Guldoman:
> something I would like to see is variables inside `" "` get highlighted a different color

Something like this maybe:
```diff
index 8f38ca8..a043eb2 100644
--- a/plugins/language_sh.lua
+++ b/plugins/language_sh.lua
@@ -7,16 +7,18 @@ local expansion_syntax = {
 }
 expansion_syntax.patterns = {
     { pattern = { "%${", "}", "\\" }, type = "keyword2", syntax = expansion_syntax },
-    { pattern = "[^%$}]+",            type = "string" },
-    { pattern = "[%$}]",              type = "string" },
+    { pattern = "$[%d$%a_@*][%w_]*",  type = "keyword2" },
+    { pattern = "[^%$}]+",            type = "string"   },
+    { pattern = "[%$}]",              type = "string"   },
 }
 
 local string_syntax = {
   name = "Shell script string",
   patterns = {
     { pattern = { "%${", "}", "\\" }, type = "keyword2", syntax = expansion_syntax },
-    { pattern = '[^%$"]+',            type = "string" },
-    { pattern = '[%$"]',              type = "string" },
+    { pattern = "$[%d$%a_@*][%w_]*",  type = "keyword2" },
+    { pattern = '[^%$"]+',            type = "string"   },
+    { pattern = '[%$"]',              type = "string"   },
   },
   symbols = {}
 }
```

We should probably also add something about `$(...)`.

Samueru-sama:
That works fine, but I had to change my color profile in lite-xl because the one I had been using doesn't differentiate the change.

Do I apply the changes as well?

>We should probably also add something about $(...).

And I just noticed that subshells don't get highlighted.

In theory it just has to match $() but not what's inside of it, my little knowledge of regexes comes from sed, grep and awk so let's see if I can do something here lol.

PR #476, opened 2024-09-02
Added Quetta

adamharrison:


Guldoman:
Looks good. Is it expected that in the manifest you're linking to different `.so` versions?

Also we should unify our branding with plugins. Officially we use `Lite XL` everywhere else.

adamharrison:
It is not! Thanks for noticing. Fixed. Also, yeah; changed.

PR #479, opened 2024-09-10
Improve langauge_kdl

Tamnac:
- fix formatting (should be all 2-space indents now)
- couple of improvements suggested by @Guldoman
- add support for node and nested multiline comments (also thanks to guldoman)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -168,7 +168,7 @@
       "id": "meta_languages",
       "mod_version": "3",
       "type": "meta",
-      "version": "0.1.18"
+      "version": "0.1.19"
     },
     {
       "description": "Align multiple carets and selections *([clip](https://user-images.githubusercontent.com/2798487/165631951-532f8d24-d596-4dd0-9d21-ff53c71ed32f.mp4))*",


Guldoman (commented):
Needs a version bump.

plugins/language_kdl.lua
```diff
@@ -3,82 +3,102 @@ local syntax = require "core.syntax"
...
+	  { pattern = "[%-+]?[%d_]+%.[%d_]+",      type = "number"   },
     { pattern = "[%-+]?[%d_]+e[%-+]?[%d_]+", type = "number"   },
     { pattern = "[%-+]?[%d_]+",              type = "number"   },
-    { pattern = "/[%-/].-\n",                type = "comment"  },
-    { pattern = {"/%*", "%*/"},              type = "comment"  },
+    { pattern = "//.*",                      type = "comment"  },
+    {
+      pattern = { "/%-.*{", "}" },
+      type = "comment", syntax = node_comment
+    },
```
Guldoman, line 51: My suggestion in #477 also covered comments like this:
```kdl
mynode /-"commented" "not commented" {
  a
  b
}
```
This doesn't, right?

Guldoman: Nope, only the single parameter should be commented.
In what cases did my suggestion not work?

PR #488, opened 2024-10-05
created transform plugin

FloppyDisco:


FloppyDisco:
i didn't see that someone was already working on  similar plugin. #138 
this one grabs the current word and supports multiple cursors.

PR #489, opened 2024-10-06
deleted lite-xl-vibe

eugenpt:
I believe the plugin is waay behind the current lite-xl (as it was developed mainly several years ago and not in any way finished at that time)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -2305,7 +2305,10 @@
       "replaces": [
         "vibe"
       ],
-      "version": "0.1"
+      "tags": [
+        "deprecated"
+      ],
+      "version": "0.2"
     },
     {
       "description": "Audio visualizer for Lite XL",


Guldoman:
You should delete it from `manifest.json`, the readme is autogenerated.

Still, @adamharrison, do we have a way to mark a plugin as deprecated? So people could still install it, but get a big warning.
Would also allow for it to be un-deprecated in the future and people would just get the updated version.

adamharrison:
I can add that in. Would that work for you @eugenpt to simply mark as deprecated?

eugenpt:
🤦🤦🤦 I missed the 'auto-generated' part... Deprecated would be fine, if it's not too much trouble

adamharrison:
Yeah; just add `deprecated` to `tags` (you can see an example of adding a tag by looking at the `language_` plugins) with this pr, and I'll merge it and make sure it basically just doesn't get added to the README (or gets added with a warning).

adamharrison:
OK, I've added in the `deprecated` tag instead. Merging.

PR #497, opened 2024-11-04
Add support for Plank's config and theme files.

PerilousBooklet:


plugins/language_ini.lua
```diff
@@ -4,7 +4,7 @@ local syntax = require "core.syntax"
 
 syntax.add {
   name = "INI",
-  files = { "%.ini$", "%.inf$", "%.cfg$", "%.editorconfig$" },
+  files = { "%.ini$", "%.inf$", "%.cfg$", "%.editorconfig$", "%.theme$", "%.dockitem$" },
   comment = ';',
   patterns = {
     { pattern = ";.-\n", type = "comment" },
```
Guldoman, line 10: I know it's unrelated, but could you change this line and the next in the usual way?

Guldoman (commented):
Needs a version bump.

plugins/language_ini.lua
PerilousBooklet: What do you mean by `usual way` ?

Guldoman: > What do you mean by `usual way` ?

`.-\n` -> `.*`

> And for the version bump, is `0.1.1` fine ? It's a minor addition.

Yep.

PR #513, opened 2025-01-20
Update Odin plugin to support more types and keywords

pgmtx:
I also removed some unused keywords.


adamharrison:
Ditto, please bump the version in manifest.json.

pgmtx:
Done.

PR #515, opened 2025-01-22
Update minimap.lua

bf:
fixes https://github.com/lite-xl/lite-xl/issues/2011

### DIFF
--- a/plugins/minimap.lua
+++ b/plugins/minimap.lua
@@ -535,7 +535,8 @@ function MiniMap:draw()
         local start = 1
         while true do
           -- find text followed spaces followed by newline
-          local s, e, w, eol = string.ufind(text, "[^%s]*()[ \t]*()\n?", start)
+          local success, s, e, w, eol = pcall(function() return string.ufind(text, "[^%s]*()[ \t]*()\n?", start) end)
+          if not success then break end
           if not s then break end
           local nchars = w - s
           start = e + 1


adamharrison:
Y'know, I'd been meaning to fix this for a while. Nice work. Can you bump the version of `minimap` in the manifest?

plugins/minimap.lua
```diff
@@ -535,7 +535,8 @@ function MiniMap:draw()
         local start = 1
         while true do
           -- find text followed spaces followed by newline
-          local s, e, w, eol = string.ufind(text, "[^%s]*()[ \t]*()\n?", start)
+          local success, s, e, w, eol = pcall(function() return string.ufind(text, "[^%s]*()[ \t]*()\n?", start) end)
+          if not success then break end
```
Guldoman, line 539: To be honest I think this could just be replaced with standard `string.find`.
In the end it's not a big deal if we consider the (mostly rare) multibyte characters as multiple characters.

Also because I'm noticing that a few lines later we're using `string.sub` with indexes returned from `string.ufind`, which is not really correct.

Guldoman (commented):
Ideally our utf-8 functions shouldn't just crash, but that's a change we need to do in Lite XL itself.

Other than the `string.find` solution, it might be a good idea to consider using `pcall` around the parent loop, so that we don't call it way too much (iirc it has some performance implications, but I'd have to check).

plugins/minimap.lua
adamharrison: This strikes me as the best move actually.

bf: Feel free to take over this PR and change it as you want

adamharrison (approved):

PR #518, opened 2025-01-22
minimap: add `always_visible` option

bf:
new option for minimap `always_visible` (default: false) which prevents minimap from hiding. user can always see minimap and does not need to hover over the scroll bar in order to show the minimap.

### DIFF
--- a/plugins/minimap.lua
+++ b/plugins/minimap.lua
@@ -25,6 +25,7 @@ local Scrollbar = require "core.scrollbar"
 -- General plugin settings
 config.plugins.minimap = common.merge({
   enabled = true,
+  always_visible = false,
   width = 100,
   instant_scroll = false,
   syntax_highlight = true,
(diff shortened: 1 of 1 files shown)


bf:
Further TODOs:

- [ ] ensure that `avoid_small_docs` still works by re-shuffling lines in `function MiniMap:is_minimap_enabled()`
- [ ] there seems to be a bug with the thumb rect height, I have experienced it with and without the modifications, so I'm 99% sure it is unrelated to this PR but would be nice for someone else to also check this

Guldoman:
> new option for minimap `always_visible` (default: false) which prevents minimap from hiding. user can always see minimap and does not need to hover over the scroll bar in order to show the minimap.

In what cases was it getting hidden?

bf:
For me minimap was alwas hidden until I hover over the right scroll bar. Then it appears but goes away immediately once mouse leaves.

Guldoman:
> For me minimap was alwas hidden until I hover over the right scroll bar. Then it appears but goes away immediately once mouse leaves.

I don't think that's standard behavior. Could you check if it's another plugin that's interfering?
What version of Lite XL are you using?

bf:
It was on 2.1.7 - but not sure I can reproduce any further. My plugins were a mess in that install.

I love lite-xl but there are some things I'd like to change/refactor. I've made a fork at https://github.com/bf/ofol to build a replacement for my personal use cases of lapce / sublime-text-4. 

I won't spam you with PRs any further and will implement my ideas there.

PR #522, opened 2025-01-24
Fix `for` keyword match in `language_sh`.

PerilousBooklet:
Fix the `for` keyword getting matched as `function` and not `keyword2` when in c++-like loop form.
To fix it, I just changed the priority of the pattern for all symbols.


adamharrison:
Version bump.

PerilousBooklet:
Done

PR #524, opened 2025-01-28
Update widgets version in manifest.json

leny100:
Updated widgets plugin version since the current one in manifest.json is causing errors

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -2405,7 +2405,7 @@
       "description": "Plugin library that provides a set of re-usable components to more easily write UI elements for your plugins",
       "id": "widget",
       "mod_version": "3",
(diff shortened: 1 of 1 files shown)


Guldoman:
What errors is it causing?
Without a version bump in the upstream project, we can't merge this.

leny100:
For example, when trying to click on "Mode:" option in Scale plugin settings I see the following error:
widgets-error
widget-error-log

When I manually updated widgets repo in lpm it is gone.

Guldoman:
Are you using Lite XL `master`?
This repo is for `2.1.7` only.

leny100:
Yes, I've built from master. 

If that is the case then I will close this PR, thanks for the help!

PR #525, opened 2025-01-30
Add language syntax reference to `language_openscad`.

PerilousBooklet:


PerilousBooklet:
Maybe I should take this opportunity to add the missing syntax references for the other languages as well.

adamharrison:
I mean, technically this should bump the version number, but because it's only a comment change, I'll allow it.

PR #527, opened 2025-02-15
Add Plugin Installer

nia-cloud-official:


adamharrison:
Hello!

So, this doesn't look super functional. The install instructions don't make sense, and I skimmed the code, and it also doesn't really make sense. I can only assume this was written by AI.

If you are a person; we appreciate you looking to contribute! But generally, you need to ensure that your code is tested and working, and has clear install instructions. AI is a great tool to do some basic research, and get basic feedback about concepts you're unfamiliar with, but it must be paired with expertise and testing to be useful. It can serve as a starting point, but you need to make sure that your code actually works before submitting it anywhere, or doing anything with it.

However, if you are looking for a plugin manager, I have good news! There are actually several you may be intested in, such as [Miq](https://github.com/TorchedSammy/Miq) or [lpm](https://github.com/lite-xl/lite-xl-plugin-manager) (and its corresponding `plugin_manager` plugin).

PR #529, opened 2025-02-19
Add support for Yuescript language

pgmtx:


adamharrison:
So, this actually installs with `lpm`, but technically I think this is against spec, as the remote on the other end doesn't have a `manifest.json`, so it's installing because `lpm` can infer things to plug legacy gaps, but for new plugins, we should probably have a manifest. Can you add one on the other end, or, alternatively, just use `url` instead, and link to the pinned raw like here: `https://codeberg.org/pgmtx/lite-plugin-yuescript/raw/commit/0b0a68c8a2066b8270b2e85d94711c2eb033410e/language_yuescript.lua`?

pgmtx:
Okay, how do you do a checksum? I assume it has nothing to do with the commit hash.

adamharrison:
Yeah; you just present the `sha256sum` of the file in question. For that file, it's `fec5394fadc4b140988ef61cdc7ccc2b5836c1bc4f527cc3b5869cad60cd79d3`

pgmtx:
Thanks, I just edited the manifest.json.

adamharrison:
Missing a comma, it's invalid json.

pgmtx:
Sorry, corrected it.

adamharrison:
So, I fucked up. Dunno how lol. But the checksum for that file is actually `9694b88c585d35ad37cdcd4278839710e9c3c64d8420538862213b56cd032911`. 

Sorry about that. Can you just change that one last thing, and I'll merge this? Last thing, I promise.

PR #532, opened 2025-02-24
Move my plugins into main repo

anthonyaxenov:
I decided to move (almost) all my repositories away from github to my own gitea instance, so I want to include my plugins in the main lite-xl plugin repository as-is, so that community could maintain them and maintainers could include them in future releases


adamharrison:
Makes sense. Merging.

PR #540, opened 2025-04-30
Remove devicons plugin from the list of plugins.

PerilousBooklet:
I decided to withdraw my custom icon fonts plugin as it's no longer in line to the standards of the more popular others and I am not going to update to the new font plugin structure.
Also, this plugin was originally meant for me, as such it contains some icon features that most would find broken (es. I color the icons following the original project icon's color, regardless of the theme currently loaded).

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -317,6 +317,27 @@
       "version": "0.1"
     },
     {
+      "description": "Icons font for the devicons plugin.",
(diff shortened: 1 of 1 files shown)


Guldoman:
Instead of deleting, could you set the `deprecated` tag?
There should be an example already in the manifest.

PR #545, opened 2025-05-30
language_groovy, language_sh: add more relevant file extensions

ildar:
language_groovy: Jenkinsfile
language_sh: APKBUILD

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1117,7 +1117,7 @@
       "tags": [
         "language"
       ],
-      "version": "0.2"
+      "version": "0.2.1"
     },
     {
       "description": "Syntax for the [Hare](https://harelang.org) programming language",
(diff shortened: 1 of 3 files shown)


xcb-xwii:
i think it would also make sense to remove fish from `language_sh`

it's not posix-compliant so the syntax really does not work for it

Guldoman:
> i think it would also make sense to remove fish from `language_sh`
> 
> it's not posix-compliant so the syntax really does not work for it

Is it better as completely non-highlighted? Hopefully someone writes a syntax for it.

ildar:
> Would you mind also bumping the version of the plugins inside manifest.json?

Ah! yes. Missed this one.

Guldoman:
Looks good, thank you!

PR #549, opened 2025-06-06
Pr/man syntax

DazFather:
Add plugin to support [man](https://it.m.wikipedia.org/wiki/Man_(Unix)) pages languages


Guldoman (commented):
Can't review properly at the moment, but this needs a version bump of `meta_languages`.

Issue #560, opened 2025-06-30
duplicate lines in language_bazel.lua ?


Alexey-T:
```
  { pattern = "-?%d+[%d%.eE_]*",  type = "number"   },  -- tested ok
  { pattern = '[%a_][%w_]*%f[(]', type = 'function' },  -- tested ok
  { pattern = "-?%d+[%d%.eE_]*",  type = "number"   },  -- tested ok
```

1st and 3rd - equal lines?

PR #561, opened 2025-06-30
Rename Bazel.

adamharrison:
Fixes https://github.com/lite-xl/lite-xl-plugins/issues/560.


manifest.json
```diff
@@ -1564,6 +1554,17 @@
       ],
       "version": "0.1"
     },
+    {
+      "description": "Syntax for [Bazel](https://bazel.build/) build tool files.",
+      "id": "language_bazel",
```
Guldoman, line 1559: The id likely needs to change too, right?

adamharrison: Whoops.

PR #562, opened 2025-07-07
Added f80 type to syntax

sirlilpanda:
Added the `f80` type in to the zig language syntax. This has been defined since zig version 0.10.1 and documentation for it can be found [here](https://ziglang.org/documentation/0.10.1/#Floats).


sirlilpanda:
bumped

PR #563, opened 2025-07-10
FIX: target pattern in `language_make`

PerilousBooklet:
The pattern that matches the `target` syntax element in a Makefile over-matches strings like `GIT_URL = https://www.kernel.org/pub/software/scm/git/git-$(GIT_VER).tar.xz`:

image

I also fixed the pattern for the `special target` syntax (es. `.PHONY:`).

### DIFF
--- a/plugins/language_make.lua
+++ b/plugins/language_make.lua
@@ -11,8 +11,8 @@ syntax.add {
     { pattern = "$[@^<%%?+|*]",           type = "keyword2" },
     { pattern = "$%(.-%)",                type = "symbol"   },
     { pattern = "%f[%w_][%d%.]+%f[^%w_]", type = "number"   },
-    { pattern = "%..*:",                  type = "keyword2" },
-    { pattern = ".*:",                    type = "function" },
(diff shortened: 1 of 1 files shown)


plugins/language_make.lua
```diff
@@ -11,8 +11,8 @@ syntax.add {
     { pattern = "$[@^<%%?+|*]",           type = "keyword2" },
     { pattern = "$%(.-%)",                type = "symbol"   },
     { pattern = "%f[%w_][%d%.]+%f[^%w_]", type = "number"   },
-    { pattern = "%..*:",                  type = "keyword2" },
-    { pattern = ".*:",                    type = "function" },
+    { pattern = "^%..*:%s",               type = "keyword2" },
+    { pattern = "^.*:%s",                 type = "function" },
```
Guldoman, line 15: Having a space after `:` doesn't seem to be required by make, so I don't think we can just ignore those cases.

Maybe we could try matching assignments?

PerilousBooklet:
Ok, great!

PR #564, opened 2025-07-12
First round of `language_make` improvements

Guldoman:
Next thing to add is `${...}`, `$(...)`.

Related to #563.

@PerilousBooklet what do you think about this? It should fix the issue you found.


PerilousBooklet:
Also wouldn't it be better to color the special targets (es. `.PHONY:`) with `operator` instead of `literal`? That's what VSCode does. Personally I would prefer to use `function`, just like with normal targets (es. `all:` or `all::`).

Guldoman:
> Also versioned numbers like `v1.2.3` should be colored. This `%f[%w_]v?[%d%.]+%f[^%w_]` should do the trick. Basically just add `v?` between `%f[%w_]` and `[%d%.]+%f[^%w_]`.

To be honest I would just remove that rule, and simply match numbers by themselves.

In general this syntax needs a lot of work, but we can at least properly highlight rules and assignments

PerilousBooklet:
Ok, great!

PerilousBooklet:
> > Also versioned numbers like `v1.2.3` should be colored. This `%f[%w_]v?[%d%.]+%f[^%w_]` should do the trick. Basically just add `v?` between `%f[%w_]` and `[%d%.]+%f[^%w_]`.
> 
> To be honest I would just remove that rule, and simply match numbers by themselves.
> 
> In general this syntax needs a lot of work, but we can at least properly highlight rules and assignments

So you would remove `{ pattern = "%f[%w_]v?[%d%.]+%f[^%w_]",                   type = "number"   },` and maybe replace it with `{ pattern = "v?[%d.?]+", type = "number" },` ?

PerilousBooklet:
Also we should add the language syntax references header comment:
```lua
-- Language Syntax References
-- https://www.gnu.org/software/make/manual/make.html
-- https://www.gnu.org/software/make/manual/html_node/Quick-Reference.html
```

PerilousBooklet:
I'm working on `{ pattern = "%$%(.-%)", type = "keyword2" },` and `{ pattern = "%$%{.-%}", type = "keyword2" },`, for `$(...)` and `${...}` .

PerilousBooklet:
Actually the `.PHONY:` target doesn't get matched with the code above.

PR #567, opened 2025-08-09
Add sel_length plugin

hellium6:
A plugin that "Displays selection length in status bar"

Supports multiple carets or selections.

Shows selection length like `sel: n` or in case multiple carets or selection `sels: n` at the bottom in the statusbar.

### DIFF
--- a/plugins/sel_length.lua
+++ b/plugins/sel_length.lua
@@ -1,0 +1,40 @@
+-- mod-version:3
+-- Selection length plugin
+-- Shows the selection length on statusbar
+local core = require "core"
+local StatusView = require "core.statusview"
+
+core.status_view:add_item({
+	predicate = StatusView.predicate_docview,
+	name = "doc:sel_length",
+	alignment = StatusView.Item.LEFT,
+	get_item = function()
+		local dv = core.active_view
+		-- To fix crash
+		-- e.g. search_ui dir selection causes crash otherwise
+		if dv.doc == nil then return {} end
+		local selection_length = 0
+		local selection_count = 0
+		-- Go through all the selections or carets in the docview
+		for _, line1, col1, line2, col2 in dv.doc:get_selections() do
+			local selection = dv.doc:get_text(line1, col1, line2, col2)
+			selection_length = selection_length + string.len(selection)
+			selection_count = selection_count + 1
+		end
+
+		-- If selection length is zero, don't bother showing any status
+		if selection_length == 0 then return {} end
+
+		-- Indicate if showing selection length for 1 selection, or more
+		local status_prefix = (selection_count <= 1 and "sel: " or "sels: ")
+
+		-- Show selection length in status
+		return {
+			status_prefix, selection_length,
+		}
+	end,
+	position = -1,
+	tooltip = "selection length",
+	separator = core.status_view.separator2
+})
+
--- a/manifest.json
+++ b/manifest.json
@@ -2196,6 +2196,13 @@
       "version": "0.1"
     },
     {
+      "description": "Displays selection length in status bar",
+      "id": "sel_length",
+      "mod_version": "3",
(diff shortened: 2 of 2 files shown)


plugins/sel_length.lua
```diff
@@ -0,0 +1,40 @@
+-- mod-version:3
+-- Selection length plugin
+-- Shows the selection length on statusbar
+local core = require "core"
+local StatusView = require "core.statusview"
+
+core.status_view:add_item({
+	predicate = StatusView.predicate_docview,
+	name = "doc:sel_length",
```
Guldoman, line 9: I'd use a different name for the initial part, so that it is obvious that it comes from a plugin.
Something like `sel_length:length`.

plugins/sel_length.lua
```diff
@@ -0,0 +1,40 @@
...
+	alignment = StatusView.Item.LEFT,
+	get_item = function()
+		local dv = core.active_view
+		-- To fix crash
+		-- e.g. search_ui dir selection causes crash otherwise
+		if dv.doc == nil then return {} end
```
Guldoman, line 15: This was happening because the predicate was `nil`. With a working predicate this won't be needed anymore.

plugins/sel_length.lua
```diff
@@ -0,0 +1,40 @@
...
+		local selection_length = 0
+		local selection_count = 0
+		-- Go through all the selections or carets in the docview
+		for _, line1, col1, line2, col2 in dv.doc:get_selections() do
+			local selection = dv.doc:get_text(line1, col1, line2, col2)
+			selection_length = selection_length + string.len(selection)
```
Guldoman, line 21: You could probably make this faster by iterating over the actual lines instead of retrieving the entire selected text, as this currently could be problematic for huge selections.

Also because I *think* this is called for each frame, but I'd have to check. So it might be a good idea to cache things (maybe using the doc with its `Doc:get_change_id()` as key).

Guldoman (commented):
Looking good, mostly just a potential performance issue.

I wonder if this should just be part of the core editor.

plugins/sel_length.lua
hellium6: Changed to `status:sel_length`

plugins/sel_length.lua
hellium6: > You could probably make this faster by iterating over the actual lines instead of retrieving the entire selected text, as this currently could be problematic for huge selections.

Not sure how this can be done. What methodology are you suggesting here? Caching selection length for line ranges?

Btw `Doc:get_selection_text` uses a similar technique. It uses `Doc:get_text` as I used here. I added some conditions similar to it.

> Also because I _think_ this is called for each frame, but I'd have to check. So it might be a good idea to cache things (maybe using the doc with its `Doc:get_change_id()` as key).

`Doc:get_change_id()` changes on each edit, not on selection change for example. Is it relevant on this plugin? Is there anything that changes on selection/cursor change? If not, what would be an optimal way to detect that a cache change is needed?

plugins/sel_length.lua
```diff
@@ -0,0 +1,43 @@
...
+		local dv = core.active_view
+		local selection_count = #dv.doc.selections/4
+		local selection_length = 0
+		-- Go through all the selections or carets in the docview
+		for _, line1, col1, line2, col2 in dv.doc:get_selections() do
+			if line1 ~= line2 or col1 ~= col2 then
+				local selection = dv.doc:get_text(line1, col1, line2, col2)
+				selection_length = selection_length + string.len(selection)
+			end
+		end
```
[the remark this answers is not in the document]
hellium6: > The selection cache gets invalidated when the document changes

That made sense. Thanks for the code. It worked pretty much as is. Just added a condition for single line selections.

PR #568, opened 2025-08-14
feat: enhance openselected.lua

yz778:
When no text is selected, create a selection with the contiguous text around the current cursor position.

### DIFF
--- a/plugins/openselected.lua
+++ b/plugins/openselected.lua
@@ -31,15 +31,43 @@ config.plugins.openselected = common.merge({
   }
 }, config.plugins.openselected)
 
+local function select_word_under_cursor(doc)
+  local line, col = doc:get_selection() -- will be cursor if no selection
+  local text = doc.lines[line]
+  local pattern = "[%s'\"]"
+
+  if not text or text == "" then return nil end
+
+  -- move left until non-word char
+  local start_col = col
+  while start_col > 1 and not text:sub(start_col - 1, start_col - 1):match(pattern) do
+    start_col = start_col - 1
+  end
+
+  -- move right until non-word char
+  local end_col = col
+  while end_col <= #text and not text:sub(end_col, end_col):match(pattern) do
+    end_col = end_col + 1
+  end
+
+  doc:set_selection(line, end_col, line, start_col)
+  return doc:get_text(doc:get_selection())
+end
+
 command.add("core.docview!", {
   ["open-selected:open-selected"] = function(dv)
     local doc = dv.doc
-    if not doc:has_selection() then
-      core.error("No text selected")
-      return
-    end
+    local text
 
-    local text = doc:get_text(doc:get_selection())
+    if not doc:has_selection() then
+      text = select_word_under_cursor(doc)
+      if not text or text == "" then
+        core.error("No text found at cursor")
+        return
+      end
+    else
+      text = doc:get_text(doc:get_selection())
+    end
 
     -- trim whitespace from the ends
     text = text:match( "^%s*(.-)%s*$" )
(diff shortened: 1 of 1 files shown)


plugins/openselected.lua
```diff
@@ -62,5 +90,5 @@ contextmenu:register("core.docview", {
 })
 
 
-keymap.add { ["ctrl+alt+o"] = "open-selected:open-selected" }
+keymap.add { ["ctrl+shift+o"] = "open-selected:open-selected" }
```
Guldoman, line 93: What's the reason for the keymap change? This will conflict with a default Lite XL shortcut.

plugins/openselected.lua
```diff
@@ -31,16 +31,44 @@ config.plugins.openselected = common.merge({
   }
 }, config.plugins.openselected)
 
+local function select_word_under_cursor(doc)
+  local line, col = doc:get_selection() -- will be cursor if no selection
+  local text = doc.lines[line]
+  local pattern = "[%s'\"]"
+
+  if not text or text == "" then return nil end
```
Guldoman, line 39: On "empty" lines, `text` will be a single character of `\n`.

plugins/openselected.lua
```diff
@@ -31,16 +31,44 @@ config.plugins.openselected = common.merge({
...
+  local start_col = col
+  while start_col > 1 and not text:sub(start_col - 1, start_col - 1):match(pattern) do
+    start_col = start_col - 1
+  end
+
+  -- move right until non-word char
+  local end_col = col
+  while end_col <= #text and not text:sub(end_col, end_col):match(pattern) do
+    end_col = end_col + 1
+  end
```
Guldoman, line 51: You could probably just use `string.find`.

Guldoman (commented):
Oh this is useful.

You should bump the version in `manifest.json`.

plugins/openselected.lua
```diff
@@ -31,16 +31,34 @@ config.plugins.openselected = common.merge({
...
 }, config.plugins.openselected)
 
+local function select_word_under_cursor(doc)
+  local line, col = doc:get_selection() -- will be cursor if no selection
+  local text = doc.lines[line]
+  local pattern = "%s'\""
+
+  if not text or text == "\n" then return nil end
+
+  local start_col = text:sub(1,col):find("[^"..pattern.."]")
```
Guldoman, line 41: If you want to use `find` for the backwards check, you'll have to continue finding until you go past `col`, then use the previous result.

plugins/openselected.lua
```diff
@@ -31,16 +31,34 @@ config.plugins.openselected = common.merge({
...
+  local end_col = col + text:sub(col, #text):find("["..pattern.."]") - 1
```
Guldoman, line 42: Instead of splitting the string, you can use the second parameter of `find` to specify where to start.

plugins/openselected.lua
```diff
@@ -31,16 +31,43 @@ config.plugins.openselected = common.merge({
...
+  local text = doc.lines[line]
+  local pattern = "%s'\""
+
+  if not text or text == "\n" then return nil end
+
+  -- if character at cursor is a boundary return nil
+  local char_at_cursor = text:sub(col, col)
+  if char_at_cursor:match("[" .. pattern .. "]") then
+    return nil
+  end
```
Guldoman, line 45: White testing, I noticed that if the cursor is at the end of the path, it doesn't open.
Is this expected?

It does open if it's at the start.

plugins/openselected.lua
```diff
@@ -31,16 +31,43 @@ config.plugins.openselected = common.merge({
   }
 }, config.plugins.openselected)
 
+local function select_word_under_cursor(doc)
+  local line, col = doc:get_selection() -- will be cursor if no selection
+  local text = doc.lines[line]
+  local pattern = "%s'\""
```
Guldoman, line 37: I wonder is space should be allowed.

Guldoman (commented):
Sorry it took a while to get back to this.

PR #584, opened 2025-12-04
feat: improve indentguide.lua

SantX27:
I find the Blockman extension for VSCode (https://marketplace.visualstudio.com/items?itemName=leodevbro.blockman) to be real handy, this is my attempt at replicating some of the functionality, by letting indentguide stack semitransparent boxes for each line. The effect can be enabled or disabled and it can be tuned, I'm not quite sure if the values I picked were right since it's VERY theme dependent.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -646,7 +646,7 @@
       "id": "indentguide",
       "mod_version": "3",
       "path": "plugins/indentguide.lua",
-      "version": "0.3"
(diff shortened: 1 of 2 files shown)


SantX27:
Forgot to add a screenshot to easily convey the stylistic changes:
immagine
This is using the default theme, with Line Background Alpha set at 20 and Line Background Brightness at 0.1

PR #586, opened 2025-12-29
Display count of characters and lines in the status bar

Titousensei:
New plugin to display count of characters and lines in the status bar.

Screenshot_2025-12-29_15-28-17


adamharrison:
Seems fine to me; can you add this to the `manifest.json` as well?

PR #593, opened 2026-03-25
Add slang shader language support

Blackoutburst:
Most of slang specs can be found here: https://github.com/shader-slang/slang/blob/master/docs/user-guide/02-conventional-features.md


plugins/language_slang.lua
```diff
@@ -0,0 +1,290 @@
...
+local syntax = require "core.syntax"
+
+syntax.add {
+  name    = "Slang",
+  files   = { "%.slang$" },
+  comment = "//",
+  block_comment = { "/*", "*/" },
+
+  patterns = {
+    { pattern = "//.-\n",                                    type = "comment" },
```
Guldoman:
```suggestion
    { pattern = "//.*",                                      type = "comment" },
```

plugins/language_slang.lua
```diff
@@ -0,0 +1,290 @@
+-- Author: Blackoutburst: https://github.com/Blackoutburst
```
Guldoman, line 1: Also add the spec link for future reference.

plugins/language_slang.lua
```diff
@@ -0,0 +1,290 @@
...
+  name    = "Slang",
+  files   = { "%.slang$" },
+  comment = "//",
+  block_comment = { "/*", "*/" },
+
+  patterns = {
+    { pattern = "//.-\n",                                    type = "comment" },
+    { pattern = { "/%*", "%*/" },                            type = "comment" },
+
+    { pattern = "%[%s*[%a_][%w_%.]*%s*[%(%]]",               type = "keyword2" },
```
Guldoman: What is this supposed to match?

Blackoutburst: Slang attributes like `[[vk::binding(0)]]`

Blackoutburst:
Removed the attribute regex, it works fine without them, added the slang specs link and fixed a small issue on binding semantics

PR #608, opened 2026-07-19
Add config options to `selectionhighlight` plugin

eleriaqueen:
Add these config options to `selectionhighlight` plugin:
- Highlight Box Color
- Highlight Box Thickness

### DIFF
--- a/plugins/selectionhighlight.lua
+++ b/plugins/selectionhighlight.lua
@@ -14,9 +14,7 @@ config.plugins.selectionhighlight = common.merge({
       description = "Color of the highlight box.",
       path = "custom_color",
       type = "color",
-      default = string.format("#%02X%02X%02X%02X",
-                              style.syntax.comment[1], style.syntax.comment[2], style.syntax.comment[3], style.syntax.comment[4]
-      )
+      default = string.format("#%02X%02X%02X%02X", table.unpack(style.syntax.comment)))
     },
     {
       label = "Highlight Box Thickness",
--- a/plugins/selectionhighlight.lua
+++ b/plugins/selectionhighlight.lua
@@ -9,7 +9,7 @@ config.plugins.selectionhighlight = common.merge({
   config_spec = {
     name = "Selection Highlight",
     {
-      label = "Highlight Box Colour",
+      label = "Highlight Box Color",
       description = "Color of the highlight box.",
       path = "custom_color",
       type = "color",
--- a/plugins/selectionhighlight.lua
+++ b/plugins/selectionhighlight.lua
@@ -5,6 +5,7 @@ local config = require "core.config"
 local DocView = require "core.docview"
 
 config.plugins.selectionhighlight = common.merge({
+  opacity = 255,
   thickness = 1,
   use_scale = true,
   config_spec = {
--- a/plugins/selectionhighlight.lua
+++ b/plugins/selectionhighlight.lua
@@ -6,6 +6,7 @@ local DocView = require "core.docview"
 
 config.plugins.selectionhighlight = common.merge({
   thickness = 1,
+  custom_color = style.syntax.comment,
   config_spec = {
     name = "Selection Highlight",
     {
--- a/plugins/selectionhighlight.lua
+++ b/plugins/selectionhighlight.lua
@@ -74,11 +74,7 @@ function DocView:draw_line_body(line, x, y)
           local x2 = x + self:get_col_x_offset(line, end_col + 1)
 
           local color = style.selectionhighlight or style.syntax.comment
-          local color_modified = {}
-          for idx, val in pairs(color) do
-             color_modified[idx] = val
-          end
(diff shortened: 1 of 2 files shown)


plugins/selectionhighlight.lua
```diff
@@ -1,16 +1,53 @@
...
+      min = 1,
+      max = 10
+    },
+    {
+      label = "Use SCALE Value",
+      description = "Increase thickness of highlight box acording to the SCALE value.",
+      path = "use_scale",
+      type = "toggle",
+      default = true
+    }
```
Guldoman: I'd remove this. Ideally SCALE should be something transparent to the user.

plugins/selectionhighlight.lua
```diff
@@ -1,16 +1,53 @@
...
+    name = "Selection Highlight",
+    {
+      label = "Highlight Box Opacity",
+      description = "Define the how opaque the highlight box is.",
+      path = "opacity",
+      type = "number",
+      default = 255,
+      min = 0,
+      max = 255
+    },
```
Guldoman, line 18: I'm not sure customizing opacity helps much here. I would add a way to customize the color, defaulting to the current theme.

plugins/selectionhighlight.lua
```diff
@@ -35,8 +72,17 @@ function DocView:draw_line_body(line, x, y)
...
           local x1 = x + self:get_col_x_offset(line, start_col)
           local x2 = x + self:get_col_x_offset(line, end_col + 1)
+
           local color = style.selectionhighlight or style.syntax.comment
-          draw_box(x1, y, x2 - x1, lh, color)
+          local color_modified = {}
+          for idx, val in pairs(color) do
+             color_modified[idx] = val
+          end
+          color_modified[4] = config.plugins.selectionhighlight.opacity
```
Guldoman: Definitely should use `ipairs` here.
Or even simpler:
```suggestion
          local color_modified = { color[1], color[2], color[3], config.plugins.selectionhighlight.opacity }
```

Guldoman (commented):
I would not expose the `SCALE` thing.
Instead of an option to change opacity I'd add support in the `config_spec` to specify a custom color.

This needs a version bump in the manifest.

plugins/selectionhighlight.lua
eleriaqueen: I removed the SCALE setting.

plugins/selectionhighlight.lua
eleriaqueen: I removed the opacity setting.

plugins/selectionhighlight.lua
```diff
@@ -1,16 +1,44 @@
...
+local common = require "core.common"
+local config = require "core.config"
 local DocView = require "core.docview"
 
+config.plugins.selectionhighlight = common.merge({
+  thickness = 1,
+  config_spec = {
+    name = "Selection Highlight",
+    {
+      label = "Highlight Box Colour",
```
Guldoman:
```suggestion
      label = "Highlight Box Color",
```
To keep consistency with the other strings.

plugins/selectionhighlight.lua
```diff
@@ -1,16 +1,44 @@
 -- mod-version:3
 local style = require "core.style"
+local common = require "core.common"
+local config = require "core.config"
 local DocView = require "core.docview"
 
+config.plugins.selectionhighlight = common.merge({
+  thickness = 1,
+  config_spec = {
```
Guldoman, line 10: The default value in the `config_spec` is only applied when the `settings` plugin is installed, so it should be defined in any case.
```suggestion
config.plugins.selectionhighlight = common.merge({
  thickness = 1,
  custom_color = style.syntax.comment,
  config_spec = {
```

plugins/selectionhighlight.lua
```diff
@@ -1,16 +1,44 @@
...
+    name = "Selection Highlight",
+    {
+      label = "Highlight Box Colour",
+      description = "Color of the highlight box.",
+      path = "custom_color",
+      type = "color",
+      default = string.format("#%02X%02X%02X%02X",
+                              style.syntax.comment[1], style.syntax.comment[2], style.syntax.comment[3], style.syntax.comment[4]
+      )
```
Guldoman: Slightly simplified:
```suggestion
      default = string.format("#%02X%02X%02X%02X", table.unpack(style.syntax.comment)))
```

eleriaqueen: Oh wow I didn't know about this `table.unpack()` function !

plugins/selectionhighlight.lua
eleriaqueen: That was a typo, I meant to use `Color` instead of `Colour`. :smile:

plugins/selectionhighlight.lua
eleriaqueen: Oh, thanks for explaining this one.

PR #611, opened 2026-08-30
fix/gitopen-for-windows: Fix GitOpen Plugin Broken On Windows

BotRandomness:
The `gitopen` plugin does not work on Windows, as Git for Windows returns path using `/` just as Unix based systems, instead of the typical `\` used for Windows paths. 

When then invoking command `Gitopen: Open From Commit`, since `PATHSEP` goes by system and when using Windows, the constructed path returned, for example `C:/user/project/src\main.c`, will be invalid and nothing happens. 

A normalize function for the path makes sure the path uses `/` no matter the system. Also added a toggle to enable/disable normalizing in `gitopen` plugin settings just in case.

### DIFF
--- a/plugins/gitopen.lua
+++ b/plugins/gitopen.lua
@@ -2,6 +2,7 @@
 local core = require "core"
 local command = require "core.command"
 local common = require "core.common"
+local config = require "core.config"
 
 local function exec(cmd)
   local proc = process.start(cmd)
--- a/plugins/gitopen.lua
+++ b/plugins/gitopen.lua
@@ -2,7 +2,22 @@
 local core = require "core"
 local command = require "core.command"
 local common = require "core.common"
+local config = require "core.config"
 
+config.plugins.gitopen = common.merge({
+  normalize_paths = true,
+
+  config_spec = {
+    name = "Git Open",
+    {
+      label = "Normalize Paths",
+      description = "Normalize Git paths by converting '\\' to '/'.",
+      path = "normalize_paths",
+      type = "toggle",
+      default = true
+    }
+  }
+}, config.plugins.gitopen)
 
 local function exec(cmd)
   local proc = process.start(cmd)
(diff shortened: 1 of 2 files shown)


plugins/gitopen.lua
```diff
@@ -15,18 +30,26 @@ local function exec(cmd)
...
 end
 
+-- Git for Windows uses / as Unix based systems
+local function normalize_path(path)
+  if config.plugins.gitopen.normalize_paths then
+    return path:gsub('\\', '/')
+  end
+
+  return path
+end
```
Guldoman: You can use `common.normalize_path`.

plugins/gitopen.lua
```diff
@@ -2,7 +2,22 @@
...
+    name = "Git Open",
+    {
+      label = "Normalize Paths",
+      description = "Normalize Git paths by converting '\\' to '/'.",
+      path = "normalize_paths",
+      type = "toggle",
+      default = true
+    }
+  }
+}, config.plugins.gitopen)
```
Guldoman: I think we can avoid having a config for this.

plugins/gitopen.lua
```diff
@@ -15,18 +30,26 @@ local function exec(cmd)
...
   local git_root = exec({"git", "rev-parse", "--show-toplevel"}):match( "^%s*(.-)%s*$" )
   local file_list_str = exec({"git", "show", "--name-only", "--pretty=format:", commit})
 
   local git_files = {}
   for str in string.gmatch(file_list_str, "([^\n]+)") do
-    git_files[git_root .. PATHSEP .. str] = true
+    git_files[normalize_path(git_root .. PATHSEP .. str)] = true
   end
 
   for dir, item in core.get_project_files() do
```
Guldoman: I wonder if we can avoid this potentially expensive loop.

The only case this could be useful would be to avoid opening files specified in `ignore_files`, but at this point I'd just filter them like here: https://github.com/lite-xl/lite-xl/blob/0af4dae0e2748d6afe7bded41c13d96f627cb520/data/core/init.lua#L397

plugins/gitopen.lua
BotRandomness: Oh I see, I didn't know `common` contained a similar function, thanks for the point out, updated!

plugins/gitopen.lua
BotRandomness: Yeah that is fair, updated to remove

plugins/gitopen.lua
BotRandomness: Yeah I was not too sure why the plugin did this initially, I agree it could be a bit expensive to loop if the project is big. Updated to filter like the example pointed out using `common.pattern_match`.

Guldoman (commented):
Looking great!

Just add a comment near the `common.match_pattern` stuff to explain why it's there and I'll merge.

Guldoman:
Thank you!

